PR #24, opened 2024-02-27
protocol update

cloudwebrtc:


Runtime/Scripts/Track.cs
```diff
@@ -43,20 +44,23 @@ public class Track : ITrack
...
         public bool Muted => _info.Muted;
-        public WeakReference<Room> Room { get; }
+        public WeakReference<Room> Room { internal set; get; }
         public WeakReference<Participant> Participant { get; }
 
         // IsOwned is true if C# owns the handle
         public bool IsOwned => Handle != null && !Handle.IsInvalid;
 
-        internal readonly FfiHandle Handle;
+        public readonly FfiHandle Handle;
```
davidzhao, line 53: I don't think we'd want to expose this?

Runtime/Scripts/VideoFrame.cs
```diff
@@ -109,50 +105,53 @@ public I420Buffer ToI420()
...
             handleId.Id = (ulong)Handle.DangerousGetHandle();
 
-            var argb = new ToARGBRequest();
-            argb.Buffer = handleId;
-            argb.DstPtr = (ulong)dst;
-            argb.DstFormat = format;
-            argb.DstStride = dstStride;
-            argb.DstWidth = width;
-            argb.DstHeight = height;
+            var toRGBA = new VideoConvertRequest();
```
davidzhao, line 134: nice!

davidzhao (commented):
Nice work!

Can you add a `.gitattributes` file and [add a line](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) to ignore the Proto files from diff? It would make managing changes a bit easier.

davidzhao (approved):

PR #30, opened 2024-04-18
Fix exceptions in events

sus-yoshikane-t:
1. Since the room's Sid was used and a `KeyNotFoundException` occurred in the `ParticipantDisconnected` event, so I fixed it to use the Participant's Sid.
1. Since TrackSubscribed was used and a `NullReferenceException` occurred in the `TrackUnsubscribed` event, so I fixed it to use `TrackUnsubscribed`.


cloudwebrtc (approved):
lgtm!

PR #38, opened 2024-06-14
try to fix audio delay.

cloudwebrtc:
Use threads to push frames to the ffi core to avoid audio frame accumulation


cloudwebrtc:
Depend on this fix to avoid inserting silent frames and causing latency accumulation.

https://github.com/livekit/rust-sdks/pull/353

PR #43, opened 2024-07-18
feat: support rendering callback for rtc video source.

cloudwebrtc:


Runtime/Scripts/RtcVideoSource.cs
```diff
@@ -121,12 +126,35 @@ public virtual void Stop()
...
         {
             while (_playing)
             {
                 yield return null;
-                ReadBuffer();
+                var textureChanged = ReadBuffer();
+
+                if(textureChanged || _texture2D == null)
+                {
+                    _texture2D = new Texture2D(_dest.width, _dest.height, TextureFormat.RGB24, false);
```
Mohelm97, line 149: You should destroy the previous texture2d, which will eat up the GPU memory.

Mohelm97:
we might need to destroy the texture in `dispose`

Mohelm97 (approved):

PR #44, opened 2024-07-18
Add unpublish track to local participant, fix local track published event

Mohelm97:
The pull request will add unpublish track to local participant, and it fixes and issue in local track published event as it search for the track but no one add it to local track dictionary


cloudwebrtc (approved):
lg!

PR #46, opened 2024-07-29
Add more method for Local/Remote Track.

cloudwebrtc:
close https://github.com/livekit/client-sdk-unity/issues/55
close https://github.com/livekit/client-sdk-unity/issues/52

Tested Platfroms:

- [x] Windows
- [x] macOS
- [x] iOS
- [x] Android

Updated ffi to v0.9.0.


burakkaraceylan:
How does muting the local track work? It seems a request is sent to the server. Does the muting happen on the server side? If so, does that mean the muted track is still being published (i.e. uploading data)?

cloudwebrtc:
Actually this request will be sent to rust core, which will shut down data upload internally, but yes, it would be more explicit to add a muted flag inside LocalTrack to prevent data send to ffi.

Cheney1112:
**After pulling the update, when calling the ‘MakeCall’ method, the following error occurs:**
“NullReferenceException: Object reference not set to an instance of an object
LiveKit.Room.OnEventReceived (LiveKit.Proto.RoomEvent e) (at Assets/client-sdk-unity-main/Runtime/Scripts/Room.cs:344)
LiveKit.Internal.FfiClient+<>c.<FFICallback>b__47_0 (System.Object resp) (at Assets/client-sdk-unity-main/Runtime/Scripts/Internal/FFIClient.cs:236)
UnityEngine.UnitySynchronizationContext+WorkRequest.Invoke () (at <d8e27b9b2bf34338bc0a9edd46a39109>:0)
UnityEngine.UnitySynchronizationContext.Exec () (at <d8e27b9b2bf34338bc0a9edd46a39109>:0)
UnityEngine.UnitySynchronizationContext.ExecuteTasks () (at <d8e27b9b2bf34338bc0a9edd46a39109>:0)
”
**Setting a breakpoint shows that the issue occurs in the ‘Room’ script at the following lines:**
”var participant = RemoteParticipants[e.TrackPublished.ParticipantIdentity];
var publication = new RemoteTrackPublication(e.TrackPublished.Publication.Info, FfiHandle.FromOwnedHandle(e.TrackPublished.Publication.Handle));
“
I hope to find a solution to this problem. Thank you.

cloudwebrtc:
Thanks for finding this bug. I believe this is because the corresponding version of ffi .so/dll has not been updated yet.

Cheney1112:
> Thanks for finding this bug. I believe this is because the corresponding version of ffi .so/dll has not been updated yet.

Thanks for the response :) When will the "ffi.so.dll" be updated, or is it possible that I didn't download the correct version?

PR #62, opened 2024-11-14
FFI v0.12.2 & RPC

bcherry:
This PR upgrades the FFI version to the latest and includes the RPC feature.

I did not adopt the majority of FFI changes but my limited testing did not suggest they are breaking changes so I believe this is safe but @theomonnom and @cloudwebrtc would know for sure.

The only other blocking issue is that the latest FFI does not build for all runtimes, which will prevent us from releasing a new version of this SDK.


Runtime/Scripts/Participant.cs
```diff
@@ -132,6 +139,108 @@ public void UpdateName(string name)
...
+            _rpcHandlers.Remove(method);
+
+            using var request = FFIBridge.Instance.NewRequest<UnregisterRpcMethodRequest>();
+            var unregisterReq = request.request;
+            unregisterReq.LocalParticipantHandle = (ulong)Handle.DangerousGetHandle();
+            unregisterReq.Method = method;
+            var resp = request.Send();
+        }
+
+        internal async void HandleRpcMethodInvocation(
```
bcherry, line 179: @cloudwebrtc @theomonnom  this is the piece I'm most unsure about. FFI triggers this, then we need to load the appropriate async user-supplied handler, execute it, then send its result back to FFI.  

I tried this with coroutines but as far as I could tell it required me to have a coroutine runner, which meant having a global singleton that might get overloaded from different calls? 

Instead for now it uses tasks with async/await but that makes it different than everything else

Is there a good way to have the user supply a coroutine that we can run without making a global singleton runner of our own? any other thoughts?

Runtime/Scripts/Internal/FFIClient.cs
```diff
@@ -122,13 +126,13 @@ private static void InitializeSdk()
             const bool captureLogs = false;
 #endif
 
-            NativeMethods.LiveKitInitialize(FFICallback, captureLogs);
+            NativeMethods.LiveKitInitialize(FFICallback, captureLogs, "unity", ""); // TODO: Get SDK version
```
bcherry, line 129: I verified this works for now but I'd love to get SDK version in. it's just not clear _where_ to get it from at runtime? Might need to inject it into a script like we do for the node SDK. any ideas @cloudwebrtc @theomonnom ?

cloudwebrtc: I found this but don't know if it works as expected
https://discussions.unity.com/t/how-get-packageinfo-by-name-and-version/762057/8

```C#
using UnityEngine;
using System.Linq;
using UnityEditor.PackageManager;

// ... within some class ...

private PackageInfo GetPackageInfo(string packageName)
{
    return UnityEditor.AssetDatabase.FindAssets("package")
        .Select(UnityEditor.AssetDatabase.GUIDToAssetPath)
            .Where(x => UnityEditor.AssetDatabase.LoadAssetAtPath<TextAsset>(x) != null)
        .Select(PackageInfo.FindForAssetPath)
            .Where(x => x != null)
        .First(x => x.name == packageName);
}
```

```
string packageName = "com.unity.cinemachine";
Debug.Log(GetPackageInfo(packageName).version);
```

bcherry: From what I saw online, `UnityEditor` isn't available at runtime on-device, only when running from the Unity Editor? Do you think that's accurate?

bcherry: gonna just punt on this

Runtime/Scripts/AudioSource.cs
```diff
@@ -69,7 +69,6 @@ public RtcAudioSource(AudioSource source, RtcAudioSourceType audioSourceType = R
             newAudioSource.Options.EchoCancellation = true;
             newAudioSource.Options.AutoGainControl = true;
             newAudioSource.Options.NoiseSuppression = true;
-            newAudioSource.EnableQueue = false;
```
cloudwebrtc, line 72: This line of code seems to need to be restored. This is a Unity-specific issue. Since there will be jitter when the Unity Audio Device captures audio, the queue needs to be closed inside ffi to avoid delay accumulation.

bcherry: This field was removed in https://github.com/livekit/rust-sdks/pull/425 - @theomonnom do you know if theres a different workaround we need to apply in Unity?

cloudwebrtc: ah, okay, theo has removed the silent frame insertion in ffi core, so this line of code can be removed

Runtime/Scripts/Internal/FFIClient.cs
cloudwebrtc: Yeah, you are right, maybe we can add a version.cs, but we must check that the version number in version.cs is consistent with the version number in package.json before releasing a new version

cloudwebrtc (approved):
lgtm!

Runtime/Scripts/Internal/FFIClient.cs
bcherry: in the node sdk we dynamically generate a version.js file as part of the build process, but i'm not sure there's a clean way to pull that off here.

cloudwebrtc: if the Unity build process can generate version.cs, I think it is a good idea.

PR #64, opened 2024-11-26
Fixes bugs with metadata and rpc

CalinR:
I found two issues when testing this in a unity project
1. _rpcHandlers needs to be protected because it is used in a class that inherits Participant
2. We need to set metadata on the participant before invoking metadataChanged otherwise we do not get the updated metadata in the callback.


cloudwebrtc (approved):
lgtm!

bcherry:
@CalinR thank you so much!

PR #68, opened 2024-12-23
Add mute request case to prevent exception

### LINKED ISSUE #67: Unable to mute/unmute local track
When calling ILocalTrack.SetMute() I get an exception:

```
Exception: Unknown request type: LiveKit.Proto.LocalTrackMuteRequest
```

holofermes:
Allow local mute / unmute local tracks. With this change tracks are effectively muted and no more audio/video data is sent to other participants in a room, however the change is not somehow propagated to the other participants in the room, ie: subscribing to a room's TrackMuted / TrackUnmuted never fires for a Unity client with this PR.

Fixes #67


cloudwebrtc (approved):
lgtm!

PR #70, opened 2024-12-23
Expose Participant.Attributes as a public field

### LINKED ISSUE #69: Access participant attributes
I would like to access a participant attributes as a KeyValuePair<string, string>.

holofermes:
Access participant attributes like so:

```
foreach (var participantAttribute in participant.Attributes)
{
    if (participantAttribute.Key == "YourKeyLabel")
}
```

I'm unsure how proper it is to expose this as-is Google.Protobuf.Collections.MapField<string, string>

Fixes #69


cloudwebrtc (approved):
lgtm!

PR #72, opened 2024-12-23
Expose LocalVideoTrack.source as a public field

### LINKED ISSUE #71: Determine a local video track source type
Currently it is not possible to retrieve a LocalVideoTrack source since the property is declared as private. Relaxing the field to be public would let me further determine a source's concrete type.

holofermes:
Expose the  LocalVideoTrack.source field as public. For example one could determine the concrete class like so:

``` c#
switch (localTrack.source)
{
    case CameraVideoSource cameraVideoSource:
        break;
    case ScreenVideoSource screenVideoSource:
        break;
    case TextureVideoSource textureVideoSource:
        break;
    case WebCameraSource webCameraSource:
        break;
}
```

Fixes #71

### DIFF
--- a/Runtime/Scripts/Track.cs
+++ b/Runtime/Scripts/Track.cs
@@ -126,12 +126,12 @@ namespace LiveKit
 
     public sealed class LocalVideoTrack : Track, ILocalTrack, IVideoTrack
     {
-        RtcVideoSource _source;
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):
lgtm!

PR #76, opened 2024-12-24
Increase video timestamping precision from milli to micro

### LINKED ISSUE #75: Video received does not play at the target framerate
When receiving a video from the Unity client, it feels like the frame rate is not consistent, and  I perceive a pattern of at-framerate and slow-framerate in a cyclical way. I tried printing out the numbers of frames received every second, by increasing a counter in a VideoStream.FrameReceived callback, and I get the following results:

```
frames received: 12
frames received: 10
frames received: 2
frames received: 9
frames received: 12
frames received: 11
frames received: 10
frames received: 13
frames received: 10
frames received: 12
frames received: 14
frames received: 11
frames received: 11
frames received: 13
frames received: 12
```

holofermes:
This PR increases the timestamp precision which is in turn used to determine whether or not to send frames down the stack.

I have confirmed this reaches playing back at the source's target framerate by printing out the number of frames received in a ...

I see that in the native library, there's logic to [assign a timestamp directly in rust](https://github.com/livekit/rust-sdks/blob/cc889daa26c2ea50258932708a24e095dcc850cc/libwebrtc/src/native/video_source.rs#L102-L108), so I tried setting the C# TimestampUs field to 0, and confirmed the results are the same as the playback rate matches the source.

Fixes #75

### DIFF
--- a/Runtime/Scripts/RtcVideoSource.cs
+++ b/Runtime/Scripts/RtcVideoSource.cs
@@ -188,7 +188,8 @@ namespace LiveKit
                 var capture = request.request;
                 capture.SourceHandle = (ulong)Handle.DangerousGetHandle();
                 capture.Rotation = GetVideoRotation();
-                capture.TimestampUs = DateTimeOffset.Now.ToUnixTimeMilliseconds();
+                var now = DateTimeOffset.UtcNow;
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):
lgtm!

PR #110, opened 2025-04-28
Fix Android build error

ladvoc:
Addresses a build error on Android caused by outdated FFI binaries. Android was removed from the platforms list in `install.py` by 9f9bfdf, leading to the FFI binaries not being kept up-to-date with new releases of the FFI server.


theomonnom (approved):

theomonnom:
I think we need to update the webrtc.jar as well

ladvoc:
There was previously an issue with the build workflow for libwebrtc, which was addressed by [livekit/rust-sdks#586](https://github.com/livekit/rust-sdks/pull/586). As a result, the latest release, webrtc-b99fd2c-6, did not include a libwebrtc.jar, so the version I have included was built from main.

cloudwebrtc (approved):

PR #117, opened 2025-05-16
Bump FFI to 0.12.24

bcherry:


ladvoc:
I’ve also updated *version.ini* and downloaded the new binaries. However, it looks like there is an issue with the FFI build workflow in the rust-sdks repo which causes *libwebrtc.jar* not to be included in the build artifact—this is required for Android. Once that is addressed, I will add it here.

PR #125, opened 2025-05-27
Audio capture latency reduction

ladvoc:
This PR aims to solve the outstanding audio capture latency issue. Recently, #122 fixed an issue related to video capture that was causing broader performance impacts, leading to high audio latency over time when both video and audio were enabled. Currently, audio samples are read using `OnAudioFilterRead`, enqueued, and then dequeued and captured via FFI on a dedicated thread. Based on the code history, it seems this might have been introduced as a fix for audio delay. With the issue from #122 fixed, this PR evaluates the performance of capturing directly from `OnAudioFilterRead` with no additional queuing on the Unity side; my initial testing on iOS and macOS shows this works well, delivering noticibly lower audio latency that does not increase over time.

I would like to discuss potential impacts of this change and test across more platforms. Also, it is worth noting since the last discussion about this in #62, the `queue_size_ms` option has been added by livekit/rust-sdks#425 but has not been adopted yet.

Related PRs and discussion in chronological order:
- #26: Initial introduction of dedicated thread for audio capture
- #29: Dedicated thread for audio capture removed
- livekit/rust-sdks#353
- #38: Re-introduction of dedicated thread, adds queueing
- livekit/rust-sdks#425: `enable_queue` option removed, replaced by `queue_size_ms`
- #62: Relevant discussion

### DIFF
--- a/Runtime/Scripts/WebCameraSource.cs
+++ b/Runtime/Scripts/WebCameraSource.cs
@@ -62,7 +62,6 @@ namespace LiveKit
                 _previewTexture.width != width ||
                 _previewTexture.height != height)
             {
-                Debug.Log("Creating new texture");
                 // Required when using Allocator.Persistent
                 if (_captureBuffer.IsCreated)
                     _captureBuffer.Dispose();


cloudwebrtc (approved):
lgtm!

PR #146, opened 2025-08-15
disposed RGBA video frames to fix indefinite linear growth of memory

s-hamdananwar:


Runtime/Scripts/VideoStream.cs
```diff
@@ -113,14 +113,15 @@ public IEnumerator Update()
                 }
                 var rgba = VideoBuffer.ToRGBA();
```
ladvoc, line 114: You might consider using [*using var*](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using) syntax here; this works since the root class of `RGBABBuffer` implements `IDisposable`.

ladvoc (approved):
LGTM, nice catch!

PR #180, opened 2026-01-08
fix: fix memory leaks for audio stream.

cloudwebrtc:


Runtime/Scripts/AudioStream.cs
```diff
@@ -96,7 +96,7 @@ private void OnAudioStreamEvent(AudioStreamEvent e)
             if (e.MessageCase != AudioStreamEvent.MessageOneofCase.FrameReceived)
                 return;
 
-            var frame = new AudioFrame(e.FrameReceived.Frame);
+            //var frame = new AudioFrame(e.FrameReceived.Frame);
```
xianshijing-lk: remove ?

Runtime/Scripts/AudioStream.cs
```diff
@@ -105,15 +105,52 @@ private void OnAudioStreamEvent(AudioStreamEvent e)
...
+                            short sample = monoSamples[i]; // mono
+                            int dstIndex = i * 2;
+
+                            stereoSamples[dstIndex] = sample; // Left
+                            stereoSamples[dstIndex + 1] = sample; // Right
+                        }
+
+                        // Cast the stereo short span to bytes and write to the buffer
+                        ReadOnlySpan<byte> stereoBytes = MemoryMarshal.AsBytes(stereoSamples);
+                        _buffer?.Write(stereoBytes);
```
xianshijing-lk: nit, probably just do 

_buffer?.Write(MemoryMarshal.AsBytes(stereoSamples));

Issue #169, opened 2025-11-10
Silent Audio Failure When Receiving Mono Audio in Unity SDK


FushimiYuki:
## Problem
When a Unity client receives **mono (1 channel)** audio from remote participants, the audio **silently fails to play** with no errors or warnings. The issue is **hard to debug** because:

- ✅ `AudioStream` creates successfully
- ✅ `AudioSource.isPlaying = true`
- ✅ All logs show "success"
- ❌ But no audio is heard

## Reproduction Steps

1. **Remote participant** publishes audio using a mono microphone (most built-in laptop mics are mono)
2. **Unity client** subscribes to the audio track:
   ```csharp
   var audioSource = gameObject.AddComponent<AudioSource>();
   var audioStream = new AudioStream(remoteAudioTrack, audioSource);
   ```
3. **Result**: Complete silence, no error messages
4. **Other clients** (Web, mobile) hear the audio fine

## Environment
- Unity 2022.3+ with default audio settings (stereo output)
- Remote participant with mono microphone
- LiveKit Unity SDK (latest)

## Root Cause

**Location**: `client-sdk-rust~/webrtc-sys/src/audio_resampler.cpp`

WebRTC's `RemixAndResample()` returns **0 bytes** when converting mono→stereo with same sample rate:

```
Input:  1 channel, 480 samples, 48kHz
Output: 2 channels, 0 samples, 48kHz  ← samples_per_channel is 0
Result: Empty data, no audio
```

The function correctly sets `num_channels=2` but fails to set `samples_per_channel`, resulting in empty output.

## Impact

- **Severity**: High - Complete audio loss
- **Silent failure**: No errors in logs, very hard to debug
- **Common scenario**: Many users have mono microphones
- **All platforms affected**: macOS, Windows, Linux, iOS, Android

## Workaround

Add this to `Runtime/Scripts/AudioStream.cs` in `OnAudioStreamEvent()`:

```csharp
var uFrame = _resampler.RemixAndResample(frame, _numChannels, _sampleRate);

// Workaround for mono→stereo bug
if ((uFrame == null || uFrame.Length == 0) && 
    frame.NumChannels == 1 && _numChannels == 2 && 
    frame.SampleRate == _sampleRate)
{
    // Manual mono to stereo conversion
    int samplesPerChannel = (int)frame.SamplesPerChannel;
    short[] monoData = new short[samplesPerChannel];
    short[] stereoData = new short[samplesPerChannel * 2];
    
[...]

PR #189, opened 2026-02-25
Mono audio tracks work in livekit client (tested in unity)

davidhughhenrymack:
Please merge so i don't have to keep a fork :)

### DIFF
--- a/Runtime/Scripts/AudioStream.cs
+++ b/Runtime/Scripts/AudioStream.cs
@@ -17,6 +17,7 @@ namespace LiveKit
         private readonly AudioSource _audioSource;
         private RingBuffer _buffer;
         private short[] _tempBuffer;
+        private short[] _upmixBuffer;
         private uint _numChannels;
         private uint _sampleRate;
         private AudioResampler _resampler = new AudioResampler();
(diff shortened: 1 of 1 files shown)


MaxHeimbrock:
Hey David, as you might have seen we have put some work into the Unity SDK since the 1.3.3 release. I just saw your PR today and wanted to check in if you are still working with the SDK and seeing the issue? The audio output is a bit more robust now.

I see you also commented on this (GitHub issue)[https://github.com/livekit/client-sdk-unity/issues/169], as you can see I was not able to repro.

davidhughhenrymack:
Thank you! We moved to android :)

MaxHeimbrock:
Hi David,

So this issue only happened on Unity iOS, but it works on Unity Android? I
did not mention the issue being platform specific. Or did you move to the
native LiveKit Android SDK without Unity?

Best regards,
Max

Runtime/Scripts/AudioStream.cs
```diff
@@ -106,11 +107,35 @@ private void OnAudioStreamEvent(AudioStreamEvent e)
...
+                        }
+
+                        var upmixed = MemoryMarshal.Cast<short, byte>(_upmixBuffer.AsSpan(0, stereoSamples));
+                        _buffer?.Write(upmixed);
+                    }
+                    else if (frame.Length > 0 && frame.NumChannels == _numChannels && frame.SampleRate == _sampleRate)
+                    {
+                        var raw = new Span<byte>(frame.Data.ToPointer(), frame.Length);
+                        _buffer?.Write(raw);
+                    }
```
xianshijing-lk, line 138: How about converting it from stereo to mono ? 

As least, I think we should log an error in the else case, and print out the values of all variables, we probably need to handle those use cases gratefully if they happen

davidhughhenrymack:
I've not been tracking since our team moved to android so i cannot answer
this, sorry about that

PR #205, opened 2026-03-26
Stress tests

MaxHeimbrock:
Measuring general connection latency, audio latency, video latency and desync.
Based on: https://github.com/livekit/client-sdk-cpp/tree/main/src/tests/stress


Tests/PlayMode/LatencyTests.cs
```diff
@@ -0,0 +1,839 @@
...
+
+        // =====================================================================
+        // Test 2: Audio Latency Measurement using Frequency-Tagged Pulses
+        // =====================================================================
+        #region Audio Latency
+
+        [UnityTest, Category("E2E")]
+        public IEnumerator AudioLatency()
+        {
+            Application.targetFrameRate = 10;
```
MaxHeimbrock: To simulate low framerates

PR #207, opened 2026-04-01
Remove discard of "stale" audio messages, catchup logic for AudioBuffer

MaxHeimbrock:
Since we merged
https://github.com/livekit/client-sdk-unity/pull/198/changes#diff-ab16de9768def77346ac0bc3e6ebc2cb95c10853f5d857fab22222025438e4da 
data can not be stale in the ring buffer due to old callbacks firing from the main dispatch queue. Callbacks are always up to date, so we can simplify some things.

Also minor improvements to underrun conditions.

I also added a catchup logic to the ring buffer. If it is above 50% fill, it will skip 5% on every read to consume faster.

=========================================

Also here is the prove that with main thread skip, data keeps flowing in when app is backgrounded:

Both measurements are already skipping the main dispatch, so data keeps coming in when backgrounded.
Left without clear, right with clear:

Screenshot 2026-04-01 at 17 03 22

### DIFF
--- a/Runtime/Scripts/AudioStream.cs
+++ b/Runtime/Scripts/AudioStream.cs
@@ -29,10 +29,6 @@ namespace LiveKit
         private const float BufferSizeSeconds = 0.2f;  // 200ms ring buffer for all platforms
         private const float PrimingThresholdSeconds = 0.03f;  // Wait for 30ms of data before playing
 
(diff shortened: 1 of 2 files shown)


xianshijing-lk (approved):
Interesting, I thought the background tasks won't run well when the apps are in the background.

PR #219, opened 2026-04-08
Improved token handling

MaxHeimbrock:
## Background:

Token usage was a pain for first time developers of LiveKit. Especially Unity devs are not always good at web technologies. I implemented a version of the TokenSource concept we have in other SDKs as well to lower the entry barrier for new devs.

## Changes:

Reuses TokenService.cs and utils from our [Agent Example](https://github.com/livekit-examples/agents-example-unity/tree/main) (thanks @ladvoc) and adopt it into TokenSource.cs.

You can create new config assets, one empty one is in the Meet project already.

## How to test:

Read the README.md paragraph about tokens and see if this makes sense.

Use the LiveKit Meet sample scene. Put an auth config instance into the TokenService:
Screenshot 2026-04-10 at 14 44 48

This is how the UI of the assets look like:
Screenshot 2026-04-10 at 14 45 34
Screenshot 2026-04-10 at 14 45 27
Screenshot 2026-04-10 at 14 45 19

### Tokens

You need a token to join a LiveKit room as a participant. Read more about tokens here: https://docs.livekit.io/frontends/reference/tokens-grants/

To help getting started with tokens, use `TokenSource.cs` with a `TokenSourceConfig` ScriptableObject (see https://docs.livekit.io/frontends/build/authentication/#tokensource). Create a config asset via **Right Click > Create > LiveKit > Token Source Config** and select one of three token source types:

#### 1. Literal
Use this to pass a pregenerated server URL and token. Generate tokens via the [LiveKit CLI](https://docs.livekit.io/frontends/build/authentication/custom/#manual-token-creation) or from your [LiveKit Cloud](https://cloud.livekit.io/) project's API key page.

#### 2. Sandbox
For development and testing. Follow the [sandbox token server guide](https://docs.livekit.io/frontends/build/authentication/sandbox-token-server/) to enable your project's sandbox and get the sandbox ID. Optional connection fields (room name, participant name, agent name, etc.) can be configured in the inspector — leave blank for server defaults.

#### 3. Endpoint
For production. Point to your own token endpoint URL and add any required authentication headers. Uses the same connection options as Sandbox. See the [endpoint token generation guide](https://docs.livekit.io/frontends/build/authentication/endpoint/).

#### Usage

Add a `TokenSource` component to a GameObject, assign your `TokenSourceConfig` asset, then fetch connection details before connecting:

```cs
var connectionDetailsTask = _tokenSource.FetchConnectionDetails();
yield return new WaitUntil(() => connectionDetailsTask.IsCompleted);

if (connectionDetailsTask.IsFaulted)
{
    Debug.LogError($"Failed to fetch connection details: {connectionDetailsTask.Exception?.InnerException?.Message}");
    yield break;
}

var details = connectionDetailsTask.Result;
_room = new Room();
var connect = _room.Connect(details.ServerUrl, details.ParticipantToken, new RoomOptions());
```


Runtime/Scripts/TokenService/TokenService.cs
```diff
@@ -0,0 +1,84 @@
...
+using UnityEngine;
+
+namespace LiveKit
+{
+    public class TokenService : MonoBehaviour
+    {
+        [SerializeField] private AuthConfig _config;
+
+        private static readonly string SandboxUrl = "https://cloud-api.livekit.io/api/sandbox/connection-details";
+        private static readonly HttpClient HttpClient = new HttpClient();
```
1egoman: Sorry, a small drive by comment - it looks like what you have here is fairly close to the existing [TokenSource](https://docs.livekit.io/frontends/build/authentication/#tokensource) concept that currently exists in other sdks. Maybe it is worth trying to make this explicitly align (ie, use the same names, expose the same set of behavior, etc)?

FWIW I don't think this would be all that hard, at least from a quick glance I think you have the majority of the required behavior already handled here, so I think it would just involve some renaming and maybe some slight reabstraction.

MaxHeimbrock: Thanks Ryan, that is a great idea. Let me see what I have to do for this to be aligned.

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -0,0 +1,114 @@
...
+                    parts.Add("\"participant_attributes\":{" + string.Join(",", attrParts) + "}");
+            }
+
+            if (!string.IsNullOrEmpty(config.AgentName) || !string.IsNullOrEmpty(config.AgentMetadata))
+            {
+                var agentParts = new List<string>();
+                AddJsonField(agentParts, "agent_name", config.AgentName);
+                AddJsonField(agentParts, "metadata", config.AgentMetadata);
+                var agentDispatch = "{" + string.Join(",", agentParts) + "}";
+                parts.Add($"\"room_config\":{{\"agents\":[{agentDispatch}]}}");
```
1egoman: This might just be because I am unfamiliar with typical unity patterns - but I would think it would be important to use a json library here instead of hand assembling the json? In particular, I suspect `AgentMetadata` could fairly often contain special characters (a fairly common case is putting an escaped json scring itself within the agent's metadata) so it would be good to make sure that won't cause unexpected breakage.

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -0,0 +1,114 @@
...
+                case AuthType.Literal:
+                    return new ConnectionDetails
+                    {
+                        server_url = _config.ServerUrl,
+                        participant_token = _config.Token
+                    };
+
+                default:
+                    throw new InvalidOperationException("Unknown auth type");
+            }
```
1egoman: Another thing where I am not familiar with typical Unity patterns, but I notied that the `AuthType.Literal` case doesn't call `FetchFromTokenServer`, so for a token source of "literal" type, this method is effectively just dead weight / somewhat of a code small.

Seeing this makes me wonder if modeling this as a class hierarchy where there's a base token source (what was done on the web / most other platforms is actually two base token sources, `TokenSourceFixed` and `TokenSourceConfigurable`, [here is the relevant code if useful](https://github.com/livekit/client-sdk-js/blob/5864b2a9e7ebc34a43e0c159bdfcfc21db8da1e7/src/room/token-source/types.ts#L48-L85)) and then higher level token sources which inherit from the base would be better? Curious what you think about this.

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace LiveKit
+{
+    public class TokenSource : MonoBehaviour
+    {
+        [SerializeField] private TokenServerConfig _config;
+
+        private static readonly string SandboxUrl = "https://cloud-api.livekit.io/api/v2/sandbox/connection-details";
+        private static readonly HttpClient HttpClient = new HttpClient();
+
+        public async Task<ConnectionDetails> FetchConnectionDetails()
```
1egoman, line 18: Similarly, on the web the decision was made to pass some of the data from `_config` here in at `fetch` / `FetchConnectionDetails` call time ([this](https://github.com/livekit/client-sdk-js/blob/5864b2a9e7ebc34a43e0c159bdfcfc21db8da1e7/src/room/token-source/types.ts#L57-L66) is what that data looks like on the web at least). 

Definitely less important and might be harder to do in a unity context if the token source is configured graphically somehow (I think that's what `TokenServerConfigEditor.cs` is about?) but wanted to call out the discrepancy at least.

1egoman (commented):
These are definitely fairly uninformed comments given my lack of unity expertise, but hopefully at least somewhat helpful!

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Yes, sadly this is something that standard Unity json can not fulfill other than this. There is a package called Newtonsoft json which is the standard .NET lib for handling json. It is an additional dependency but a very standard one.

I agree that I would rather like to have this clean with the dependency.

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: I had a look and the way it is set up in Unity is I think ideal for the usability. This way we only have one type of scriptable object, which is easiest for UI in the Unity inspector. Claude Code agrees as well:

` The JS split exists because the JS SDK passes options at call time (tokenSource.fetch({ roomName: "..." })), whereas in Unity the options are baked into the ScriptableObject inspector. My recommendation: the split doesn't add value in Unity's inspector-driven model`

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks. Yes I had a look and I think you never really create the config at runtime, but have them prepared graphically in the Editor. This is a Unity specific flow and should look like it is.

Runtime/Scripts/TokenSource/TokenSource.cs
1egoman: Sounds good. If it's easy I would suggest testing these config values with some special characters (things that would break string encoding, like strings containing `"`s / etc) just to make sure it's resilient. I'm in particular a bit worried because of the tendency for agent metadata to be an encoded json string like `"{\\"foo\\":\\"bar\\"}"`.

1egoman: Ah, I see you ended up using the json library instead. Assuming that new dependency is ok, that works even better!

1egoman (approved):
Makes sense to me at a high level! If you want a deeper review of your c# code it might be worth waiting for a review from the other reviewers.

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Trying to build an example I realized that you are right and there are cases where you want to drive the TokenSource at runtime with different options, instead of using prepared assets.

This is the new PR: https://github.com/livekit/client-sdk-unity/pull/256

PR #221, opened 2026-04-08
SuperPR: Simplification of code

MaxHeimbrock:
### Background

We might have duplicated code, too convoluted classes, etc. I wanted to see what we can find and fix with Claude

### Changes
                                                                                                                                         
  - Replace Inject() 165-line type-dispatch switch with cached reflection — new request types added to the proto no longer require manual switch-case additions. Uses the same reflection + caching pattern     
  already established by InitializeRequestAsyncId in the same file.
  - Replace EnsureClean() null-check chains with protobuf oneof check — two methods (130 lines combined) that manually checked 30+ fields each are replaced by a single 2-line method using MessageCase != None.
   The FfiRequest overload was removed entirely as it had no call sites.                                                                                                                                        
  - Consolidate 4 identical instruction classes into generic FfiInstruction<TCallback> — SetLocalMetadataInstruction, SetLocalNameInstruction, SetLocalAttributesInstruction, and UnpublishTrackInstruction were
   character-for-character identical. A single generic class replaces all of them. Complex instruction classes (e.g., PublishTrackInstruction, PerformRpcInstruction) are unchanged.                            
  - Minor fixes: mark _isDisposed as volatile for cross-thread visibility, fix garbled comment in FFIClient.cs, fix non-atomic pool creation race in ThreadSafeMultiPool.
                                                                                                                                                                                                                
  Breaking changes                                          
                                                                                                                                                                                                                
  Return types of SetMetadata(), SetName(), SetAttributes(), and UnpublishTrack() on LocalParticipant changed from concrete instruction types to FfiInstruction<T>. Callers using yield return or var are       
  unaffected since the base type (YieldInstruction) is unchanged.

### DIFF
--- a/Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
+++ b/Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Threading;
 using System.Runtime.CompilerServices;
 using System.Reflection;
@@ -11,6 +12,7 @@ namespace LiveKit.Internal.FFIClients
     {
         private static long nextRequestAsyncId;
         private static readonly ConcurrentDictionary<Type, Action<object, ulong>?> requestAsyncIdSetters = new();
+        private static readonly ConcurrentDictionary<Type, Action<FfiRequest, object>> injectors = new();
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static ulong InitializeRequestAsyncId<T>(this T request)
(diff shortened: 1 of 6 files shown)


Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
```diff
@@ -71,303 +73,43 @@ public static ulong InitializeRequestAsyncId<T>(this T request)
...
 
-                // Video
-                request.NewVideoStream != null
-                || request.NewVideoSource != null
-                || request.CaptureVideoFrame != null
-                || request.VideoConvert != null
-                ||
+            var injector = injectors.GetOrAdd(request.GetType(), static type =>
+            {
+                const BindingFlags flags = BindingFlags.Instance | BindingFlags.Public;
```
MaxHeimbrock, line 94: Using reflection here might reduce performance, we have to be careful.

MaxHeimbrock:
Split into smaller PRs

PR #223, opened 2026-04-09
Fix null participant in ActiveSpeakersChanged list

MaxHeimbrock:
### Background

GetParticipant() can return null if the identity is unknown. The speakers list was adding nulls without checking, which could cause NullReferenceException in event handlers.

### Changes

What did you do?

- Check for null


MaxHeimbrock:

PR #224, opened 2026-04-09
Fix NullReferenceException in RTCConfiguration.ToProto()

MaxHeimbrock:
### Background

IceServers field was never initialized, so calling ToProto() without explicitly setting it would throw when iterating.

### Changes

What did you do?

- Init IceServers array with empty array.


MaxHeimbrock:

PR #225, opened 2026-04-09
Fix null dereference in TrackSubscribed handler

MaxHeimbrock:
### Background

The code checked if publication was null, then immediately accessed publication.Sid — causing a NullReferenceException. The block was also logically inverted (adding a null to the dictionary). The publication should already exist from a prior TrackPublished event, so the dead block is removed.


MaxHeimbrock:

PR #228, opened 2026-04-09
Guard against KeyNotFoundException in room event dispatch

MaxHeimbrock:
### Background

`Room.OnEventReceived()` uses direct dictionary indexing (`RemoteParticipants[key]`, `participant.Tracks[key]`) in 8+ event handlers. If the server sends an event for an unknown participant or track — which can happen during reconnection races or with late-arriving events — these throw `KeyNotFoundException` and crash the entire event dispatch loop.

### Changes

- Replace direct dictionary access with `TryGetValue` + early return in 8 event handlers: `ParticipantDisconnected`, `TrackPublished`, `TrackUnpublished`, `TrackSubscribed`, `TrackUnsubscribed`, `TrackMuted`, `TrackUnmuted`, `ConnectionQualityChanged`
- Add null checks for `GetParticipant()` return values in `TrackMuted`, `TrackUnmuted`, and `ConnectionQualityChanged`
- Log debug messages for skipped events to aid troubleshooting
- Remove dead null-dereference block in `TrackSubscribed` (inverted null check that would crash)

### DIFF
--- a/Runtime/Scripts/Room.cs
+++ b/Runtime/Scripts/Room.cs
@@ -326,14 +326,22 @@ namespace LiveKit
                 case RoomEvent.MessageOneofCase.ParticipantDisconnected:
                     {
                         var sid = e.ParticipantDisconnected.ParticipantIdentity;
-                        var participant = RemoteParticipants[sid];
+                        if (!_participants.TryGetValue(sid, out var participant))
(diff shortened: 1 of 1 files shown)


MaxHeimbrock:

PR #230, opened 2026-04-09
Code Scan Fix: Replace manual EnsureClean null-checks with protobuf MessageCase

MaxHeimbrock:
## Summary

- The two `EnsureClean` extension methods manually checked 30+ fields out of 75 cases for null to verify a protobuf oneof was empty. Protobuf's built-in `MessageCase` property already tracks which oneof field is set, making all those null checks redundant.
- The `FfiRequest` overload had zero call sites and is removed entirely.
- The `FfiResponse` overload (called from `Pools.cs`) is replaced with a single `MessageCase != None` check that catches all cases, which also provides a better error message by including the offending case name.
- Adds a test which failed before the fix comes in with the second commit: https://github.com/livekit/client-sdk-unity/actions/runs/24729492636

## Test plan

- [ ] Any FFI request that goes through the response pool exercises the new `EnsureClean` path (connect, publish track, etc.)
- [ ] CI passes (existing edit-mode and play-mode tests cover the pool return path)


Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
```diff
@@ -238,136 +238,11 @@ public static void Inject<T>(this FfiRequest ffiRequest, T request)
...
-                || response.LocalDataTrackUnpublish != null
-                || response.LocalDataTrackIsPublished != null
-                || response.SubscribeDataTrack != null
-                || response.RemoteDataTrackIsPublished != null
-                || response.DataTrackStreamRead != null
-            )
-            {
-                throw new InvalidOperationException("Response is not cleared: ");
-            }
+            if (response.MessageCase != FfiResponse.MessageOneofCase.None)
```
MaxHeimbrock: Lets write a test against this behaviour

ladvoc (approved):
LGTM ✅, this is much cleaner.

PR #234, opened 2026-04-10
Improve test reliability and prevent CI timeouts

MaxHeimbrock:
### Background

Tests were intermittently timing out on CI with no way to diagnose which test was hanging. The root causes were data tracks tests that had race conditions. In Unity's batch mode timeout attributes of tests don't work.

### Changes

- Make DataTracks tests more robust against race condition with retry loop, this was otherwise causing flakyness and timeouts!
- Replace `Time.time` with `Time.realtimeSinceStartup` in `Expectation.Wait()` for reliable timeouts in batch mode
- Add a health-check step in CI that polls the LiveKit server before running tests
- Normalize assertion patterns to use idiomatic `Assert.IsNull`/`Assert.IsNotNull` instead of `if/Assert.Fail`
- Enable `RoomSid_StartsWithRM` test (previously ignored, now passing)
- Fix `Connect_FailsWithInvalidUrl` test by ignoring the potential LiveKit error log with `LogAssert.Expect`
- Moved some DataTracks tests over to EditMode tests, as they were [Test] instead of [Unity Test]

### More Background on Timeouts:

Locally, a Unity PlayMode test [Unity Test] is timing out per default after 180s. You can change the time out with the NUnit attribute `[Timeout(15000)]`. Locally this works, on CI it is sadly not, as proven here: https://github.com/livekit/client-sdk-unity/pull/238.

If a test hangs on CI, Github has a 10 minute time out right now. But that means, we don't get any results what and why timed out. This is how it looks like if Github has to shut down the Unity tests:

Screenshot 2026-04-13 at 17 55 51

That is why all coroutines we await now have their own time out.


Scripts~/run_unity_tests.sh
```diff
@@ -3,19 +3,19 @@
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 
-# Auto-detect Unity: pick the latest installed version from Unity Hub
+# Auto-detect Unity: pick the oldest installed version from Unity Hub
```
MaxHeimbrock, line 6: I have both Unity 2023 and Unity 6 installed, but want to use 2023 by default.

cloudwebrtc (approved):
lgtm!

PR #244, opened 2026-04-16
CI: add build-ffi job if rust sdk changed

MaxHeimbrock:
## Summary
- Add a Linux-only `build-ffi` job that rebuilds `liblivekit_ffi.so` from the `client-sdk-rust~` submodule when the PR touches it (mirrors the upstream Rust SDK's `ffi-builds.yml` Linux build)
- Upload the rebuilt `.so` as an artifact; the `test` job downloads it before running so tests exercise the current submodule rather than the LFS-tracked binary
- When the submodule is untouched, the build step is skipped and tests use the committed binary as before

See the mechanism working here:
- Without Rust change: job is skipped - test fails: 
  - https://github.com/livekit/client-sdk-unity/actions/runs/24449650951/job/71435050881
- With Rust change: job is executed - test is fixed: 
  - https://github.com/livekit/client-sdk-unity/actions/runs/24449863960/job/71436846364

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,55 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
+  build-ffi:
+    name: Rebuild FFI (Linux)
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lukasIO (approved):
nice!

PR #248, opened 2026-04-17
Batch 1 Bugfix: NullReferenceException in RTCConfiguration.ToProto and IceServer.ToProto

MaxHeimbrock:
## Summary
Default-constructed instances of \`RTCConfiguration\` and \`IceServer\` threw on \`ToProto()\` because their array fields (\`IceServers\`, \`Urls\`) have no initializer and the proto serialization path iterated them without a null guard.

- \`RTCConfiguration.ToProto()\` → \`NullReferenceException\` at \`Room.cs:75\` (foreach over null array)
- \`IceServer.ToProto()\` → \`ArgumentNullException\` from \`Google.Protobuf.Collections.RepeatedField<string>.AddRange(null)\` at \`Room.cs:36\`

Both are now null-guarded. Adds three regression tests in \`Tests/EditMode/RoomOptionsTests.cs\`:
- \`RTCConfiguration_ToProto_DefaultInstance_DoesNotThrow\`
- \`IceServer_ToProto_WithNullUrls_DoesNotThrow\`
- \`IceServer_ToProto_WithAllFieldsSet_Succeeds\` (happy path, validates serialized output)

\`Tests/EditMode/LiveKit.EditModeTests.asmdef\` now references \`Google.Protobuf.dll\` so the happy-path test can inspect the \`RepeatedField<string>\` the proto exposes.

## Test plan
- [x] \`./Scripts~/run_unity.sh test -m EditMode -f LiveKit.EditModeTests.RoomOptionsTests\` → 3/3 pass
- [x] Full EditMode regression: \`./Scripts~/run_unity.sh test -m EditMode\` → 35 passed, 2 skipped (pre-existing), 0 failed

### DIFF
--- a/Runtime/Scripts/Room.cs
+++ b/Runtime/Scripts/Room.cs
@@ -33,7 +33,8 @@ namespace LiveKit
             var proto = new Proto.IceServer();
             proto.Username = Username;
             proto.Password = Password;
-            proto.Urls.AddRange(Urls);
+            if (Urls != null)
+                proto.Urls.AddRange(Urls);
 
             return proto;
         }
@@ -72,9 +73,12 @@ namespace LiveKit
                     break;
             }
 
-            foreach (var item in IceServers)
+            if (IceServers != null)
             {
-                proto.IceServers.Add(item.ToProto());
+                foreach (var item in IceServers)
+                {
+                    proto.IceServers.Add(item.ToProto());
+                }
             }
 
             return proto;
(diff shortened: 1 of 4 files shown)


Runtime/Scripts/Room.cs
```diff
@@ -33,7 +33,8 @@ public Proto.IceServer ToProto()
             var proto = new Proto.IceServer();
             proto.Username = Username;
             proto.Password = Password;
-            proto.Urls.AddRange(Urls);
+            if (Urls != null)
```
MaxHeimbrock, line 36: Lets revisit if we can make this code path private or internal or remove completely if unused in our examples.

MaxHeimbrock: Also ask DevSup if we ever get requests for custom ICE servers

MaxHeimbrock: Also remove ContinualGatheringPolicy?

PR #249, opened 2026-04-17
Batch 1: EditMode tests

MaxHeimbrock:
Some simple EditMode tests generated together with Claude to increase test coverage.

### DIFF
--- a/Tests/EditMode/DataTrackTests.cs
+++ b/Tests/EditMode/DataTrackTests.cs
@@ -30,5 +30,38 @@ namespace LiveKit.EditModeTests
             Assert.IsNotNull(duration);
             Assert.GreaterOrEqual(duration.Value, 0.0);
         }
+
+        [Test]
+        public void DataTrackFrame_WithUserTimestampNow_DoesNotMutateOriginal()
+        {
+            var original = new DataTrackFrame(new byte[] { 0x01 });
+            Assert.IsNull(original.UserTimestamp);
+
+            var stamped = original.WithUserTimestampNow();
+
+            Assert.IsNull(original.UserTimestamp);
+            Assert.IsNotNull(stamped.UserTimestamp);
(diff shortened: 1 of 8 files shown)


Tests/EditMode/DataTrackTests.cs
```diff
@@ -30,5 +30,38 @@ public void DataTrackFrame_DurationSinceTimestamp_ReturnsValue()
             Assert.IsNotNull(duration);
             Assert.GreaterOrEqual(duration.Value, 0.0);
         }
+
+        [Test]
+        public void DataTrackFrame_WithUserTimestampNow_DoesNotMutateOriginal()
```
ladvoc, line 35: praise: Good to test these behaviors.

Tests/EditMode/VideoFrameTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using LiveKit.Proto;
+using NUnit.Framework;
+
+namespace LiveKit.EditModeTests
+{
+    public class VideoFrameTests
+    {
+        static VideoBufferInfo BufferInfo(VideoBufferType type)
+        {
+            return new VideoBufferInfo { Type = type, Width = 16, Height = 16 };
```
ladvoc, line 12: question: What happens if a zero or negative width/height is passed when creating the frame buffer?

ladvoc (approved):
LGTM ✅

Tests/EditMode/VideoFrameTests.cs
MaxHeimbrock: In production, the VideoBufferInfo is created either from the local video source (webcam etc) or from the remote video track. I think if we were to have corrupted width and height in there, we are doing something wrong earlier and most likely some marshaling will fail. The function is internal so these should actually never created by hand.

PR #250, opened 2026-04-19
Batch 2 Bugfix: IRemoteTrack.SetEnabled: dispatch EnableRemoteTrackRequest over FFI

MaxHeimbrock:
## Summary

\`IRemoteTrack.SetEnabled\` was completely non-functional:

\`\`\`
System.Exception: Unknown request type: LiveKit.Proto.EnableRemoteTrackRequest
  at LiveKit.Internal.FFIClients.FfiRequestExtensions.Inject[T] (...) in FfiRequestExtensions.cs:237
  at LiveKit.Internal.FFIClients.Requests.FfiRequestWrap\`1[T].Send () (...) in FfiRequestWrap.cs:59
  at LiveKit.IRemoteTrack.SetEnabled (System.Boolean enabled) (...) in Track.cs:59
\`\`\`

The proto field \`EnableRemoteTrack\` exists on \`FfiRequest\` but \`FfiRequestExtensions.Inject\`'s type switch had no case for \`EnableRemoteTrackRequest\`, so it fell through to \`default\` and threw. Adds the one missing case (matches the pattern used for all other request types).

## Regression test

\`Tests/PlayMode/RemoteTrackEnabledRegressionTests.cs\` — publishes an audio track from a second room, waits for \`TrackSubscribed\` on the subscriber, and calls \`SetEnabled(false)\` / \`SetEnabled(true)\`. Before the fix this test throws; after, it round-trips through FFI cleanly.

To make the test runnable in batch-mode CI (no real microphone), also introduces \`Tests/PlayMode/Utils/SineWaveAudioSource.cs\` — an \`RtcAudioSource\` subclass that pushes 20ms 48kHz PCM frames from a \`System.Threading.Timer\`, avoiding Unity's audio subsystem. Same utility is also part of an in-flight coverage expansion branch (\`max/more-tests/batch-2\`); merging either first is fine, content is identical.

## Test plan
- [x] \`./Scripts~/run_unity.sh test -m PlayMode -f LiveKit.PlayModeTests.RemoteTrackEnabledRegressionTests -n 5\` → 5/5 iterations pass

### DIFF
--- a/Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
+++ b/Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
@@ -152,6 +152,9 @@ namespace LiveKit.Internal.FFIClients
                 case LocalTrackMuteRequest localTrackMuteRequest:
                     ffiRequest.LocalTrackMute = localTrackMuteRequest;
                     break;
+                case EnableRemoteTrackRequest enableRemoteTrackRequest:
+                    ffiRequest.EnableRemoteTrack = enableRemoteTrackRequest;
+                    break;
                 case E2eeRequest e2EeRequest:
                     ffiRequest.E2Ee = e2EeRequest;
                     break;
--- a/Tests/PlayMode/RemoteTrackEnabledRegressionTests.cs
+++ b/Tests/PlayMode/RemoteTrackEnabledRegressionTests.cs
@@ -6,13 +6,11 @@ using UnityEngine.TestTools;
 
 namespace LiveKit.PlayModeTests
 {
-    /// <summary>
-    /// Regression test for IRemoteTrack.SetEnabled: EnableRemoteTrackRequest was missing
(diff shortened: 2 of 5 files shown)


Runtime/Scripts/Internal/FFIClients/FfiRequestExtensions.cs
```diff
@@ -152,6 +152,9 @@ public static void Inject<T>(this FfiRequest ffiRequest, T request)
                 case LocalTrackMuteRequest localTrackMuteRequest:
                     ffiRequest.LocalTrackMute = localTrackMuteRequest;
                     break;
+                case EnableRemoteTrackRequest enableRemoteTrackRequest:
```
ladvoc, line 155: comment (non-blocking): From our sync, it might be worth later revisiting how the FFI client is implemented to reduce the possibility for this kind of bug.

ladvoc (approved):
LGTM ✅

PR #254, opened 2026-04-20
Batch 4: FFI boundary & memory coverage

MaxHeimbrock:
Adds 5 tests covering FFI boundary behaviors that previously had no coverage:

EditMode (2 tests, Tests/EditMode/RoomDisconnectTests.cs):
- Disconnect on a never-connected Room (no FFI handle) does not throw
- Disconnect called twice on an unconnected Room is idempotent

PlayMode (3 tests, Tests/PlayMode/FfiBoundaryTests.cs):
- Disconnect called twice on a connected Room is idempotent
- PerformRpc with a 14 KiB ASCII payload round-trips correctly (tests large-payload marshaling in both directions: request payload and response string)
- Three concurrent PerformRpc invocations dispatched without yielding between them all complete with their own distinct payloads, exercising the FFI request memory pool and confirming request_async_id uniqueness


Tests/PlayMode/FfiBoundaryTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            yield return context.ConnectAll();
+            Assert.IsNull(context.ConnectionError, context.ConnectionError);
+
+            var room = context.Rooms[0];
+            Assert.DoesNotThrow(() => room.Disconnect(), "first Disconnect threw");
+            Assert.DoesNotThrow(() => room.Disconnect(), "second Disconnect threw");
+        }
+
+        [UnityTest, Category("E2E")]
+        public IEnumerator PerformRpc_NearLimitPayload_EchoesRoundTrip()
```
ladvoc: question: Should this go in a separate RPC test suite?

ladvoc (approved):
LGTM ✅

Tests/PlayMode/FfiBoundaryTests.cs
MaxHeimbrock: Agreed, will do that now.

cloudwebrtc (approved):

PR #256, opened 2026-04-21
TokenSource config at runtime via options

MaxHeimbrock:
### Background

In the first implementation of TokenSource, there was only the option to use a preconfigured config created as a ScriptableObject in the editor. Often you want to create the options at runtime (e.g. set participant name at runtime)

### Changes

- Adds TokenSourceFetchOptions like https://github.com/livekit/client-sdk-js/blob/5864b2a9e7ebc34a43e0c159bdfcfc21db8da1e7/src/room/token-source/types.ts#L57-L66
- Old path still valid


Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -42,9 +49,9 @@ public async Task<ConnectionDetails> FetchConnectionDetails()
             }
         }
 
-        private async Task<ConnectionDetails> FetchFromTokenSource(string url, IEnumerable<StringPair> headers)
+        private async Task<ConnectionDetails> FetchFromTokenSource(string url, IEnumerable<StringPair> headers, TokenSourceFetchOptions options)
```
1egoman: One thing worth considering - in the js token source implementation (and in I think all other implementations as well), we've modeled two types of token sources, a "configurable" token source and a "fixed" token source. The former takes options as input which configures the generated token output and the later doesn't take any options / just returns a token derived independently. 

Relevant js types: https://github.com/livekit/client-sdk-js/blob/67fee1b275aefc14b656318ab910bc5bb8d93e95/src/room/token-source/types.ts#L48-L85

Since there is just one class here for all token source types and there isn't a `TokenSourceLiteral` (the main "fixed" implementer), I'm not sure if this same pattern could be fully established here. But since you are adding `options` now, I do sort of wonder if there's any provisions you can make to the interface to allow for compatibility for both token source patterns in the future.

1egoman (approved):

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Oh yeah right I remember, I think you brought it up before. I think this could also come up in the future, so I will see how to add the literal as well at runtime.

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -15,7 +15,14 @@ public class TokenSource : MonoBehaviour
         private static readonly string SandboxUrl = "https://cloud-api.livekit.io/api/v2/sandbox/connection-details";
         private static readonly HttpClient HttpClient = new HttpClient();
 
-        public async Task<ConnectionDetails> FetchConnectionDetails()
+        public Task<ConnectionDetails> FetchConnectionDetails() => FetchConnectionDetails(null);
```
xianshijing-lk: [question](https://pullpo.io/cc?l=question) 
curiously, is this an API breaking change ?

Runtime/Scripts/TokenSource/TokenSourceFetchOptions.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+
+namespace LiveKit
+{
+    public class TokenSourceFetchOptions
+    {
+        public string RoomName { get; set; }
```
xianshijing-lk: nit, can it the value be changed after the initialization ? 
I am wondering if it makes more sense to do things like: 
public string RoomName { get; init; }

Runtime/Scripts/TokenSource/TokenSource.cs
xianshijing-lk: nit, I know all unity C# object is nullable by default, but it is still slightly better to be explicit, maybe we should consider: 
TokenSourceFetchOptions? options

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -69,35 +76,45 @@ private async Task<ConnectionDetails> FetchFromTokenSource(string url, IEnumerab
...
-        private static TokenSourceRequest BuildRequest(TokenSourceConfig config)
+        private static TokenSourceRequest BuildRequest(TokenSourceConfig config, TokenSourceFetchOptions options)
         {
             var request = new TokenSourceRequest
             {
-                RoomName = NullIfEmpty(config.RoomName),
-                ParticipantName = NullIfEmpty(config.ParticipantName),
-                ParticipantIdentity = NullIfEmpty(config.ParticipantIdentity),
-                ParticipantMetadata = NullIfEmpty(config.ParticipantMetadata),
+                RoomName = NullIfEmpty(options?.RoomName ?? config.RoomName),
```
xianshijing-lk: [question](https://pullpo.io/cc?l=question)

can options.RoomName be "" ? probably that is a edge case that we need to handle anyway ? 

as example, when
options.RoomName = ""
config.RoomName = "abc"

the current code will return null, is it the expected result ? 
or we should do thing like 
NullIfEmpty(options?.RoomName) ?? NullIfEmpty(config.RoomName)

xianshijing-lk (commented):
lgtm after you address the comments.

xianshijing-lk (approved):

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: ~It is not, the drop in async does not matter to the caller since it is a direct forward of the inner async method.~

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: I agree, this needs some helper code for NullIfEmpty but then the asymmetry is gone. Added it in a commit.

Runtime/Scripts/TokenSource/TokenSourceFetchOptions.cs
MaxHeimbrock: Yes, this is possible if we add a helper we already had in the AgentExample, now that I remember (in newer C# than Unity uses you don't need the helper): https://github.com/livekit-examples/agents-example-unity/blob/main/AgentsExample/Assets/Scripts/Support/IsExternalInit.cs

Then we can use `init;` like here: https://github.com/livekit-examples/agents-example-unity/blob/main/AgentsExample/Assets/Scripts/TokenService.cs#L62

I updated the PR to have init on all TokenSourceFetchOptions fields.

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using UnityEngine;
+
+namespace LiveKit
+{
+    public class TokenSourceHelper : MonoBehaviour
```
MaxHeimbrock: This was previously called TokenSource and was the "UI" way of doing things, described in the original PR description: https://github.com/livekit/client-sdk-unity/pull/219

Renaming it is of course a breaking change, but the API has not changed.

Samples~/Meet/Assets/Runtime/MeetManager.cs
```diff
@@ -15,7 +15,7 @@
 /// <summary>
 /// Manages a LiveKit room connection with local/remote audio and video tracks.
 /// </summary>
-[RequireComponent(typeof(TokenSource))]
+[RequireComponent(typeof(TokenSourceHelper))]
```
MaxHeimbrock: In MeetManager we can see that the naming changed but API is the same.

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: I tried my best reusing the structure from JS for runtime creation and usage of the TokenSource concept and then renamed the preexisting UI helper to `TokenSourceHelper`. I think this is ready for another round of review.

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: ~It is now, renamed `TokenSource` to `TokenSourceHelper`~

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -4,53 +4,105 @@
...
+            _endpointUrl = endpointUrl;
+            _headers = headers;
         }
 
-        private async Task<ConnectionDetails> FetchFromTokenSource(string url, IEnumerable<StringPair> headers)
+        public async Task<ConnectionDetails> FetchConnectionDetails(TokenSourceFetchOptions options)
         {
-            var requestBody = BuildRequest(_config);
+            var requestBody = BuildRequest(options);
             var jsonBody = JsonConvert.SerializeObject(requestBody);
```
1egoman, line 100: suggestion: You may consider adding some caching logic in here - if the `TokenSourceFetchOptions` hasn't changed and the token is still valid it should be able to be reused.

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
1egoman: I don't have a sense of how tolerable breaking changes are here because this is not yet a "production" sdk, so I think it would be good to confirm generally that this is ok.

I also wonder if there's maybe a better name for `TokenSourceHelper` which helps make it clear that the `TokenSourceConfig` is configured in the UI - maybe something like `TokenSourceInspector`? But I think you have a better beat on what the average unity developer would prefer.

Runtime/Scripts/TokenSource/TokenSourceComponent.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            switch (_tokenSource)
+            {
+                case ITokenSourceConfigurable configurable:
+                    return await configurable.FetchConnectionDetails(Coalesce(_config, options));
+        
+                case ITokenSourceFixed fixedSource:
+                    if (options != null)
+                        Debug.Log("Token source helper has fixed config, fetch options are ignored.");
+                    return await fixedSource.FetchConnectionDetails();
```
1egoman, line 69: thought: I think this is a little weird - the `TokenSourceHelper` sort of _sometimes_ is an `ITokenSourceFixed` and sometimes an `ITokenSourceConfigurable`. I don't know if I see an obviously better way to fix this though so - the thing that immediately came to mind was like a `TokenSourceHelperFixed` / `TokenSourceHelperConfigurable` type of split but the user shouldn't really have to care about the difference. `TokenSourceHelper<Fixed>` / `TokenSourceHelper<Configurable>` has the same problem. So maybe this is just an inevitable downside to this single `TokenSourceHelper` approach.

question: Just for my own understanding, do you have access to the currently configured subtype of `TokenSourceConfig` while the user is editing the c# code in unity? Or do all you know is that "the config is a subtype of TokenSourceConfig and that's it". If it's the former, maybe there's something fancy that could be done here.

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+                    if (options != null)
+                        Debug.Log("Token source helper has fixed config, fetch options are ignored.");
+                    return await fixedSource.FetchConnectionDetails();
+
+                default:
+                    throw new InvalidOperationException("Unknown token source type");
+            }                                                                                          
+        }       
+
+        private static TokenSourceFetchOptions Coalesce(TokenSourceConfig config, TokenSourceFetchOptions options)
```
1egoman: nitpick: Maybe rename `TokenSourceConfig` -> `TokenSourceHelperConfig`? The thing I am worried about is `TokenSourceConfig` is fairly close to `TokenSourceConfigurable`.

1egoman: Just generally though - this coalesce method is a really clever way to make this interface work, and I think is also really easy to understand!

README.md
```diff
@@ -223,7 +223,7 @@ Use the samples of the package to see how to use the SDK.
 
 You need a token to join a LiveKit room as a participant. Read more about tokens here: https://docs.livekit.io/frontends/reference/tokens-grants/
 
-To help getting started with tokens, use `TokenSource.cs` with a `TokenSourceConfig` ScriptableObject (see https://docs.livekit.io/frontends/build/authentication/#tokensource). Create a config asset via **Right Click > Create > LiveKit > Token Source Config** and select one of three token source types:
+To help getting started with tokens, use `TokenSourceHelper.cs` with a `TokenSourceConfig` ScriptableObject (see https://docs.livekit.io/frontends/build/authentication/#tokensource). Create a config asset via **Right Click > Create > LiveKit > Token Source Config** and select one of three token source types:
```
1egoman: You might consider making a docs.livekit.io update to [the section on token sources](https://docs.livekit.io/frontends/build/authentication/) to add in unity once this is merged and in a good state.

1egoman (approved):
Nice work with this!

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: I think this is a great idea, but I will push that into a follow up PR to get this one in first.

README.md
MaxHeimbrock: You mean as a code snippet right? Like this one for sandbox:

Screenshot 2026-04-30 at 13 32 44


That is a good point, I created a ticket to go over all snippets and see if I can add Unity specific ones to the tabs.

Runtime/Scripts/TokenSource/TokenSourceComponent.cs
MaxHeimbrock: Yes, that is the ugliest part of the design right now. I think from a user perspective it is great to have one UI place where you can drag and drop / file select the config you want to use. I do that myself a lot.

My solution was the `Debug.Log("Token source helper has fixed config, fetch options are ignored.");` but I also forgot to promote it to a Debug.LogWarning. This would make you aware of the footgun.

I feel like the alternatives to split up the `TokenSourceHelper` into two is creating more UI chaos and the fixed TokenSource config is probably rarely used.

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
MaxHeimbrock: Done, renamed to `TokenSourceComponentConfig`

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Done

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: It is now, renamed `TokenSource` to `TokenSourceComponent`

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
MaxHeimbrock: Thanks :)

Runtime/Scripts/TokenSource/TokenSourceHelper.cs
MaxHeimbrock: Yes I was also considering other names. In Unity, the classes inheriting from MonoBehaviour are called components as it is an Entity <=> Component system.

I renamed it to `TokenSourceComponent`, after ideating with Claude.

Runtime/Scripts/TokenSource/TokenSourceComponent.cs
MaxHeimbrock: I did not come up with a better solution right now. I think I will leave it as it is.

README.md
1egoman: Yep, exactly. I've made quite a few docs updates lately so if you can't figure it out on your own let me know and I can point you in the right direction.

Runtime/Scripts/TokenSource/TokenSourceComponent.cs
1egoman: Yep, makes sense. I couldn't come up with something better either so I think this is probably good enough as is.

PR #259, opened 2026-04-22
Fix for incremental reader chunk-drop

MaxHeimbrock:
## Summary
- Add a single EditMode regression test that pins down the chunk-drop bug in `ReadIncrementalInstructionBase<T>`.
- The base class stores each `OnChunk` call into a single `_latestChunk` field, so when several `ChunkReceived` FFI events are dispatched on the same sync-context flush (common: multiple chunks arriving between Unity frames), earlier chunks are silently overwritten before the consumer coroutine can resume from its yield.
- No Runtime changes in this PR — the queue-backed fix lands on a follow-up branch once this failing test is in place.

## Test plan
- [x] Add just the new unit test without the fix and see the CI tests fail:
https://github.com/livekit/client-sdk-unity/actions/runs/24781354985?pr=259

- [x] Fix bug in 2nd commit will make this test go green without changing the test itself.

### DIFF
--- a/Tests/EditMode/DataStreamIncrementalReadTests.cs
+++ b/Tests/EditMode/DataStreamIncrementalReadTests.cs
@@ -1,12 +1,11 @@
 using System;
 using System.Collections.Generic;
-using LiveKit;
 using LiveKit.Internal;
 using NUnit.Framework;
 
 namespace LiveKit.EditModeTests
 {
-    public class DataStreamIncrementalReadTests
+    public class DataStreamTests
     {
         private sealed class TestIncrementalReader : ReadIncrementalInstructionBase<string>
         {
--- a/Tests/EditMode/DataStreamIncrementalReadTests.cs
+++ b/Tests/EditMode/DataStreamIncrementalReadTests.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+using LiveKit;
+using LiveKit.Internal;
+using NUnit.Framework;
+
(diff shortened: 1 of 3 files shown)


Tests/EditMode/DataStreamIncrementalReadTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            public TestIncrementalReader(FfiHandle h) : base(h) { }
+            public void PushChunk(string content) => OnChunk(content);
+            public string Value => LatestChunk;
+        }
+
+        [Test]
+        public void OnChunk_MultipleChunksBeforeConsumerDrains_AllChunksAreObserved()
+        {
+            using var handle = new FfiHandle(IntPtr.Zero);
```
ladvoc, line 20: question: I'm probably missing something obvious, but how does creating a zero `FfiHandle` work here?

ladvoc (approved):
LGTM ✅

Tests/EditMode/DataStreamIncrementalReadTests.cs
MaxHeimbrock: It is a null pointer and we just need it to satisfy the constructor. In the TestIncrementalReader we can push chunks to mock callbacks without needing any native stuff, so the null handle is never used. Does that answer your question?

MaxHeimbrock: And on cleanup we check if the handle is valid, so it would also not be released.

PR #260, opened 2026-04-22
Fix AudioStream catch-up clicks: add cooldown + crossfade

MaxHeimbrock:
## Summary

- AudioStream's drift-correction path was firing on almost every audio callback and skipping samples via `RingBuffer.SkipRead`, which is a raw pointer move. The resulting step discontinuity produces a click/pop train under sustained tones.
- Add a cooldown (`SkipCooldownCallbacks = 10`) so we never skip back-to-back — that is what turns a single correction into an audible artifact train.
- When the skip does fire, read a 128-frame (~2.7 ms @ 48 kHz) post-skip window and linearly crossfade it into the tail of the output. The seam becomes a short linear ramp instead of a step — inaudible on voice and music. Guarded with `AvailableRead() >= skipBytes + crossfadeBytes` so we never trade a catch-up click for an underrun click.

Scoped change: one file (`Runtime/Scripts/AudioStream.cs`), +49/-10. No public API change, no RingBuffer/Resampler changes.

## Test plan

- [ ] Build in Unity 6000.0.49f1 (and 2023.2.20f1 if convenient) — must compile clean.
- [ ] Run `Samples~/Basic` against `livekit-server --dev` with a sustained-tone or music publisher. Before the fix: click train within seconds. After: no perceptible artifacts over a multi-minute session.
- [ ] Temporarily log `_buffer.AvailableRead()` — fill should oscillate in the 30–160 ms band, not park at 100 ms, and should not pin at the HWM.
- [ ] Confirm "AudioStream primed" appears once and no repeated underrun re-prime messages during steady-state playback.
- [ ] Background / foreground cycle still recovers cleanly (existing `OnApplicationPause` path unchanged).


Runtime/Scripts/AudioStream.cs
xianshijing-lk, line 30: I wonder if we should bump up the buffer size to 250ms for now ?  to make sure we can handle the drift better ?

Runtime/Scripts/AudioStream.cs
```diff
@@ -29,9 +30,15 @@ public sealed class AudioStream : IDisposable
...
         private const float PrimingThresholdSeconds = 0.03f;  // Wait for 30ms of data before playing
 
-        // Drift correction: skip samples when buffer fills up due to clock drift
-        private const float HighWaterMarkPercent = 0.50f;    // Target 50% fill level after correction
-        private const float SkipPerCallbackPercent = 0.05f;  // Skip 5% of callback samples per call
+        // Drift correction: skip samples when the buffer fills up due to clock drift.
+        // HWM at 50% (100ms of 200ms) so normal network jitter does not trip catch-up.
+        // Cooldown prevents back-to-back skips, which sound like a gravelly click train;
+        // one occasional skip is inaudible thanks to the crossfade in OnAudioRead.
+        private const float HighWaterMarkPercent = 0.50f;
```
xianshijing-lk, line 37: it is still 50% for the HIghWatermarkPercent ?

Runtime/Scripts/AudioStream.cs
```diff
@@ -29,9 +30,15 @@ public sealed class AudioStream : IDisposable
...
+        private const float SkipPerCallbackPercent = 0.05f;
```
xianshijing-lk, line 38: I think we'd better re-visit the logic, does skipping 5% of audio noticable ? 

if it is noticeable, we'd better re-consider the approach. Like will it work better if we just skip initial 50ms of audio in the buffer so that we will have a continuous audio stream later ?  or should we increase the buffer size ?

xianshijing-lk (approved):
lgtm, we can ship the current version though it is slightly more complicated than I would expect.

PR #266, opened 2026-04-29
Dispatch log batches off the main thread

MaxHeimbrock:
Logs from Rust were going through SynchronizationContext.Post like every other event, which is wasteful: UnityEngine.Debug.unityLogger is thread-safe, log batches are SDK-internal (no user callback expectation), and in LK_VERBOSE builds they can flood the post queue.

Forward them inline from the FFI callback alongside AudioStreamEvent and remove the now-unreachable Logs case from DispatchEvent.


MaxHeimbrock:
Done in:
https://github.com/livekit/client-sdk-unity/pull/269

PR #267, opened 2026-04-29
Cleanup Audio Dispatch

MaxHeimbrock:
### Background

Audio dispatch is handled separately and happens above the main thread dispatch and switch case:
```
if (response.MessageCase == FfiEvent.MessageOneofCase.AudioStreamEvent)
{
  Instance.AudioStreamEventReceived?.Invoke(response.AudioStreamEvent!);
  return;
}
```

### Changes

- Remove case FfiEvent.MessageOneofCase.AudioStreamEvent from switch case


Runtime/Scripts/Internal/FFIClient.cs
MaxHeimbrock, line 284: This is where we already handle the audio case.

MaxHeimbrock:
Done in:
https://github.com/livekit/client-sdk-unity/pull/269

PR #271, opened 2026-05-05
Remove Unity resampling

MaxHeimbrock:
## Summary

Remove second resampling in Unity, by configuring the Rust audio stream with correct `(channels, sampleRate)` to begin with. The problem is, that in Unity you don't know the channel number of each audio source until it runs, so we need to delay the initialization of the native audio stream until audio is read in Unity. The same lazy-create path is reused to **recreate the native audio stream when Unity's delivered format changes mid-stream** — for example after a system audio device switch.

## Changes

- **Drop C# `RemixAndResample` per-frame path.** `OnAudioStreamEvent` now just memcpys the inbound `AudioFrame` into the ring buffer — no FFI roundtrip on the OS-audio-scheduled callback thread. Rust's `NativeAudioStream` does the rate/channel conversion internally, once, with its existing pipeline.
- **Lazily create the FFI native stream from the first `OnAudioRead`.** `AudioSettings.GetConfiguration().speakerMode` disagrees with what `OnAudioFilterRead` actually delivers when the AudioSource is routed differently (per-source mono routing, spatializers, mixer chains, headphone profiles) — observed on macOS Editor + Sony headphones (system Stereo, source mono → low-pitched playback). The first callback is the authoritative source for `(channels, sampleRate)`; we post a one-shot `CreateOrRecreateFfiStream(...)` to the main thread via `FfiClient._context`, build the `NewAudioStreamRequest` with the observed values, and only then subscribe to `AudioStreamEventReceived`.
- **Recreate the native stream when Unity's delivered format changes.** A single gate in `OnAudioRead` covers both first-create and runtime mismatch (`channels != _ffiNumChannels || sampleRate != _ffiSampleRate`). On mismatch we post the same `CreateOrRecreateFfiStream` to the main thread; the new stream is built outside the lock, then under the lock we swap `_handle`, update the `_ffi*` trackers, clear the buffer, reset priming, and clear the `_pendingFfiRequest` flag. Old handle is disposed AFTER the swap so its Rust task exits cleanly. `OnAudioStreamEvent` drops frames while `_pendingFfiRequest` is set so old-format bytes don't land in the new buffer; the existing handle-id filter catches any in-flight stragglers from the old stream after the swap. This pairs with #274 (re-attach `AudioProbe` and re-`Play()` the AudioSource on device change) so audio resumes seamlessly when the user plugs/unplugs headphones, switches Bluetooth devices, etc.

## Files

- `Runtime/Scripts/AudioStream.cs` — `_resampler` field gone; `Handle` is now `_handle` behind a property; `CreateOrRecreateFfiStream` added; `_ffiNumChannels`/`_ffiSampleRate`/`_pendingFfiRequest` track FFI-side configuration; constructor only sets up AudioSource/AudioProbe/pause hook.
- `Tests/EditMode/MediaStreamLifetimeTests.cs` — assertions updated for the renamed/removed members.

## Profiler

Before change, after 5 minutes and multiple audio sources: **Total Audio CPU** reaches ~100%
Screenshot 2026-05-06 at 14 47 09

________________________________________________________________________________

With change, after 5 minutes and multiple audio sources: **Total Audio CPU** stays below ~10%
Screenshot 2026-05-06 at 14 46 48

### DIFF
--- a/Runtime/Scripts/AudioStream.cs
+++ b/Runtime/Scripts/AudioStream.cs
@@ -13,7 +13,14 @@ namespace LiveKit
     /// </summary>
     public sealed class AudioStream : IDisposable
     {
-        internal readonly FfiHandle Handle;
+        // FFI native stream is created lazily on the first OnAudioRead so we can pass
+        // Unity's actual delivered (channels, sampleRate) — not a system-speaker-mode
+        // guess. _handle is null until CreateFfiStream completes on the main thread.
+        private FfiHandle _handle;
+        internal FfiHandle Handle => _handle;
+        private readonly ulong _trackHandleId;
+        private bool _ffiCreatePosted;
+
         private readonly AudioSource _audioSource;
         private AudioProbe _probe;
         private RingBuffer _buffer;
@@ -21,7 +28,6 @@ namespace LiveKit
         private short[] _crossfadeScratch;
         private uint _numChannels;
         private uint _sampleRate;
-        private AudioResampler _resampler = new AudioResampler();
         private readonly object _lock = new object();
         private bool _disposed = false;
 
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/AudioStream.cs
```diff
@@ -73,6 +71,45 @@ public AudioStream(RemoteAudioTrack audioTrack, AudioSource source)
...
             // Subscribe to application pause events to handle background/foreground transitions
             MonoBehaviourContext.OnApplicationPauseEvent += OnApplicationPause;
+
+            // FFI stream creation is deferred to the first OnAudioRead. We subscribe to
+            // AudioStreamEventReceived only after the stream exists (CreateFfiStream).
+        }
+
+        // Called on the main thread (posted from OnAudioRead via FfiClient._context) once
+        // we know Unity's actual playback (channels, sampleRate). Builds the native stream
+        // request so Rust delivers frames already at the right format — no resample on this
```
MaxHeimbrock: Reduce comments

xianshijing-lk (approved):

PR #274, opened 2026-05-07
Recover AudioStream when system audio output device changes

MaxHeimbrock:
## Summary
- Fixes `AudioStream` going silent on macOS / Android after the system audio output device changes (e.g. unplugging headphones), where the only previous workaround was disconnecting and re-subscribing the participant.
- Subscribes to `AudioSettings.OnAudioConfigurationChanged` and, in the handler, destroys the existing `AudioProbe` + `AddComponent`s a fresh one and `Stop()`/`Play()`s the `AudioSource`. The ring buffer is cleared and re-primed so we don't drain stale frames buffered during the device-change blackout.

## Why re-`Play()` alone isn't enough
Unity tears down its native audio engine on device changes (macOS / AAudio). `AudioSource`s are stopped, but more importantly the `OnAudioFilterRead` registration of `AudioProbe` lives on the *DSP graph* and is detached during the rebuild. Unity only re-registers filter components added *after* the rebuild — so we have to recreate the `AudioProbe` to get the filter slot back. iOS doesn't hit this because route changes there don't rebuild the engine.

## Test plan
- [x] macOS Editor: unplug/plug headphones during a call — audio resumes on the new device without disconnecting the participant.
- [x] Android: same flow, with both wired and Bluetooth route changes.
- [x] iOS: regression check — route changes still work as before (no behavior change expected).
- [x] EditMode tests pass (`MediaStreamLifetimeTests` updated for the new unsubscribe).

### DIFF
--- a/Runtime/Scripts/AudioStream.cs
+++ b/Runtime/Scripts/AudioStream.cs
@@ -15,7 +15,7 @@ namespace LiveKit
     {
         internal readonly FfiHandle Handle;
         private readonly AudioSource _audioSource;
-        private readonly AudioProbe _probe;
+        private AudioProbe _probe;
         private RingBuffer _buffer;
         private short[] _tempBuffer;
         private short[] _crossfadeScratch;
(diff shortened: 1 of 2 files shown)


cloudwebrtc (approved):
lgtm!

PR #281, opened 2026-05-13
Meet: Android foreground service to survive screen lock

MaxHeimbrock:
## Summary

On Android, locking the screen / backgrounding the Meet app for ~50s causes the LiveKit SDK to exhaust its 10 reconnect attempts and emit `failed to reconnect to the livekit room`, leaving the room permanently disconnected. iOS suspends the whole process so its tokio reconnect loop is paused mid-flight and recovers on resume — Android does not get that for free; with Doze restricting the network, the loop runs at full speed against a dead socket and gives up.

This PR makes the Meet sample run a foreground service while a call is active so the OS keeps the process out of Doze and the network stays alive. iOS / desktop / editor are unaffected (call-site `#if UNITY_ANDROID && !UNITY_EDITOR` guards).

Changes:
- `Samples~/Meet/Assets/Plugins/Android/AndroidManifest.xml` — adds `FOREGROUND_SERVICE` + `FOREGROUND_SERVICE_MEDIA_PLAYBACK` permissions and a `mediaPlayback` `<service>` entry.
- `Samples~/Meet/Assets/Plugins/Android/LiveKitForegroundService.java` — small foreground service that posts a low-importance ongoing notification.
- `Samples~/Meet/Assets/Runtime/AndroidBackgroundService.cs` — thin Android-only C# bridge with `Start()` / `Stop()`.
- `MeetManager.cs` — starts the service on a successful `Connect`, stops it on `OnEndCall` / `OnDestroy`.

## Test plan

- [x] Android device: connect to a room, lock the screen, wait > 1 minute, unlock — call still alive, no `restarting connection… attempt: N` log spam.
- [x] Android device: swipe app away — participant leaves the room within the LiveKit server timeout (~15–60s). This was already the pre-existing behaviour on `main` and is server-side timeout, not regressed by this PR.
- [x] iOS device: behaviour unchanged.
- [x] Editor / desktop: behaviour unchanged.

## Notes

- `POST_NOTIFICATIONS` is intentionally **not** declared — the foreground service runs without it; only the notification visibility is gated by it, and we never request it at runtime.
- The Java service is shipped at sample level rather than inside the SDK; if we'd rather have it ship to all consumers, it can move into the package.

### DIFF
--- a/Samples~/Meet/Assets/Plugins/Android/LiveKitForegroundService.java
+++ b/Samples~/Meet/Assets/Plugins/Android/LiveKitForegroundService.java
@@ -1,0 +1,75 @@
+package io.livekit.unity.sample;
+
+import android.app.Notification;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.app.Service;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
+import android.os.IBinder;
+
+public class LiveKitForegroundService extends Service {
+
+    private static final String CHANNEL_ID = "livekit_call";
+    private static final int NOTIFICATION_ID = 1701;
+
+    public static void start(Context context) {
+        Intent intent = new Intent(context, LiveKitForegroundService.class);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            context.startForegroundService(intent);
+        } else {
+            context.startService(intent);
+        }
+    }
+
+    public static void stop(Context context) {
+        Intent intent = new Intent(context, LiveKitForegroundService.class);
+        context.stopService(intent);
+    }
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        ensureChannel();
+        startForeground(NOTIFICATION_ID, buildNotification());
+    }
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+        return START_STICKY;
+    }
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return null;
+    }
+
+    private void ensureChannel() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) return;
+        NotificationManager nm = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+        if (nm == null) return;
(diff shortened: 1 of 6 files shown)


Samples~/Meet/Assets/Runtime/MeetManager.cs
```diff
@@ -90,6 +90,9 @@ private void OnDestroy()
         }
         CleanUpAllTracks();
         _webCamTexture?.Stop();
+#if UNITY_ANDROID && !UNITY_EDITOR
```
SavvyBeny, line 93: Not everybody uses MeetManager. Does this mean every client has to use this on their own?

Samples~/Meet/Assets/Plugins/Android/LiveKitForegroundService.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private Notification buildNotification() {
+        Notification.Builder builder;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            builder = new Notification.Builder(this, CHANNEL_ID);
+        } else {
+            builder = new Notification.Builder(this);
+        }
+        return builder
```
SavvyBeny, line 68: I guess this is hardcoded for now just for testing?

SavvyBeny (approved):

Samples~/Meet/Assets/Plugins/Android/LiveKitForegroundService.java
MaxHeimbrock: Yes this was an initial try how this fix behaves in practice. I am investigating other options as well, since I am worried this one could drain the battery if you lock your screen.

In this draft PR it is also pretty hard coded. If this is the best solution, I would want this to go into the SDK and not much work for the client code.

Samples~/Meet/Assets/Runtime/MeetManager.cs
MaxHeimbrock: I think this is a decision still to be made. Depending on the downsides of this solution, it might be a client decision if this behaviour is wanted. 

If there are no downsides though, I would rather want to make it the generic behaviour.

PR #282, opened 2026-05-15
Send ReadyForRoomEvent after Connect to flush queued room events

MaxHeimbrock:
## Dependency

Depends on: https://github.com/livekit/rust-sdks/pull/1068 - have to check why this even passes one of the two test suites without.

Should be reviewed once I bump the Rust FFI to the new release 0.12.59 that will have the depending change.

## Summary

- `Room.OnConnect` now sends the new FFI `ReadyForRoomEventRequest` after subscribing to `FfiClient.RoomEventReceived` and before firing the public `Connected` event, completing the new ready-handshake the Rust SDK requires.
- Adds a PlayMode regression test (`LateJoinTrackSubscriptionTests`) that publishes 2 audio + 2 video tracks first, then connects a late joiner with `TrackSubscribed` pre-wired, and asserts every snapshot publication produces exactly one `TrackSubscribed` event.
- Regenerates the C# proto bindings (the new request/response, plus mechanical drift from running protoc 34 against the current Rust protos) and patches `DataTrack.cs` for the `DataTrackStreamEOS.error` field type change.

## Why
Without `ReadyForRoomEventRequest`, room events emitted between `ConnectCallback` and the C# listener registration could be silently dropped — late joiners would miss `TrackSubscribed` / `ParticipantConnected` for state that already existed at connect time. The new Rust contract (on `client-sdk-rust` branch `ladvoc/ffi-event-flush`) makes this deterministic: the connect task parks for 1 s waiting for `ReadyForRoomEventRequest` and disconnects the room with `ConnectionTimeout` if it never arrives. This PR ports that contract to the Unity SDK.


xianshijing-lk (approved):
I tested it locally. I am landing it to avoid breaking the functionalities

PR #285, opened 2026-05-19
Max: Integrate the PlatformAudio to unity

MaxHeimbrock:
### Background

Why are you doing this change? For example:

Token generation was too hard, so ...

### Changes

What did you do?

- Added new mechanism to generate tokens based on sandbox server id
- ...


xianshijing-lk (commented):
from a quick look, it is very good, one question about an api

PR #288, opened 2026-05-20
Improve Android fix branch

MaxHeimbrock:
### Background

iOS does not offer to chose the output device via the WebRTC ADM. We have to skip the selection entirely it seems.

```
Routing is controlled by AVAudioSession:

Skip SetRecordingDevice / SetPlayoutDevice entirely. iOS only ever has one logical WebRTC device, and the
actual mic/speaker selection is done by AVAudioSession (which IOSAudioSessionHelper.LiveKit_ConfigureAudioSessionForVoIP() already configures in
your constructor)
```

Also on Android asks for microphone permission before starting the PlatformAudio

### Changes

- Skip SetRecordingDevice / SetPlayoutDevice entirely
- Add missing inject for `SetRecordingDevice` and `SetPlayoutDevice`
- StartRecording is now a coroutine first asking for microphone permission on Android

### DIFF
--- a/Runtime/Scripts/PlatformAudio.cs
+++ b/Runtime/Scripts/PlatformAudio.cs
@@ -202,6 +202,7 @@ namespace LiveKit
                 throw new InvalidOperationException($"Recording device index {index} out of range (max: {recording.Count - 1})");
 
             var deviceId = recording[(int)index].Guid;
+
             // Note: On Android, devices don't have GUIDs - they're identified by index only.
             // Android also only reports a single "default" microphone because the system
             // automatically selects the best input source based on the audio mode.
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/PlatformAudio.cs
```diff
@@ -4,6 +4,7 @@
 using LiveKit.Proto;
 using LiveKit.Internal;
 using LiveKit.Internal.FFIClients.Requests;
+using System.Diagnostics;
```
MaxHeimbrock: Can be removed

Runtime/Scripts/PlatformAudio.cs
```diff
@@ -286,6 +293,13 @@ public void SetPlayoutDevice(uint index)
         /// </exception>
         public void SetPlayoutDevice(string deviceId)
         {
+#if UNITY_IOS && !UNITY_EDITOR
```
xianshijing-lk, line 296: should we do it for both IOS and Android ?

PR #289, opened 2026-05-21
Stage 1: Add GetAwaiter to YieldInstruction + StreamYieldInstruction

MaxHeimbrock:
## Summary
- Adds `GetAwaiter()` to `YieldInstruction` and `StreamYieldInstruction` so callers can `await room.Connect(...)` and similar one-shot/streaming FFI ops without taking on a UniTask dependency.
- Continuation is invoked from the existing `IsDone` / `IsCurrentReadDone` / `IsEos` property setters — every existing concrete instruction (Connect, PublishTrack, PerformRpc, SendText/File, stream open/write/close, etc.) becomes awaitable with no change to its completion path.
- Race between FFI-thread completion and main-thread await registration is resolved with a sentinel-value `Interlocked.CompareExchange` on a single continuation slot. Reset() clears the slot so each chunk of a stream gets its own awaiter.
- `GetResult()` is a no-op so the `await` surface keeps strict parity with `yield return` (callers still inspect `IsError`). A throwing variant can be layered on later if we want it.

This is Stage 1 of the staged UniTask migration described in `Logs~/unitask-investigation/unitask-migration-analysis.md`. It is non-breaking — no public API changes, no new dependencies, all coroutine call sites untouched. If you like the shape, Stage 2 layers an optional UniTask surface behind a scripting define.

## Test plan
- [x] `Scripts~/run_unity.sh build macos` — clean compile.
- [x] `Scripts~/run_unity.sh test -m PlayMode -f Connect_FailsWithInvalidUrl_Awaitable` — new test passes; awaiter resumes and observes `IsError`.
- [x] `Scripts~/run_unity.sh test -m PlayMode -f 'Connect_FailsWithInvalidUrl$'` — original coroutine-path test still passes (no regression).
- [x] Reviewer: spot-check `await room.Connect(...)` in a sample once a live `livekit-server --dev` is running, to confirm main-thread resumption.

### DIFF
--- a/Runtime/Scripts/Internal/YieldInstruction.cs
+++ b/Runtime/Scripts/Internal/YieldInstruction.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using UnityEngine;
 
 namespace LiveKit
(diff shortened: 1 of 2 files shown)


MaxHeimbrock:
I was able to convert `OnStartCall` to async void and then `ConnectToRoom()` to async Task with awaiter instead of yield. 

So this worked well.

PR #293, opened 2026-06-08
Platform audio for Unity

MaxHeimbrock:
### Background

Integrates the Rust Platform Audio Module. The old Unity audio input and output is still supported, the platform audio is now a second path to get audio input and output.

### Changes

- Adds platform audio APIs
- Adds choice in Meet sample to either use platform audio or Unity audio

### DIFF
--- a/Runtime/Plugins/iOS/LiveKitAudioSession.mm
+++ b/Runtime/Plugins/iOS/LiveKitAudioSession.mm
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2024 LiveKit, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <AVFoundation/AVFoundation.h>
+
+extern "C" {
+
+/// Configures the iOS audio session for VoIP/WebRTC use.
+/// This sets AVAudioSessionCategoryPlayAndRecord with VoiceChat mode,
+/// which enables the VPIO (Voice Processing IO) AudioUnit for:
+/// - Hardware echo cancellation (AEC)
+/// - Automatic gain control (AGC)
+/// - Noise suppression (NS)
+///
+/// Call this before creating PlatformAudio to ensure WebRTC can
+/// properly initialize the microphone and speaker.
+void LiveKit_ConfigureAudioSessionForVoIP() {
+    AVAudioSession* session = [AVAudioSession sharedInstance];
+    NSError* error = nil;
+
+    // Configure for VoIP with echo cancellation
+    BOOL success = [session setCategory:AVAudioSessionCategoryPlayAndRecord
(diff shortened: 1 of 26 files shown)


Runtime/Scripts/PlatformAudio.cs
```diff
@@ -0,0 +1,380 @@
...
+                    Guid = device.HasGuid ? device.Guid : null
+                });
+            }
+
+            return (recording, playout);
+        }
+
+        /// <summary>
+        /// Sets the recording device (microphone) by index.
+        ///
```
MaxHeimbrock, line 183: Add disclaimer here as well about mobile platform no-ops

Runtime/Scripts/PlatformAudio.cs
```diff
@@ -0,0 +1,380 @@
...
+    /// var track = LocalAudioTrack.CreateAudioTrack("microphone", source, room);
+    ///
+    /// // Publish track
+    /// await room.LocalParticipant.PublishTrack(track, options);
+    ///
+    /// // Dispose when done
+    /// platformAudio.Dispose();
+    /// </code>
+    /// </example>
+    public sealed class PlatformAudio : IDisposable
```
MaxHeimbrock, line 71: Maybe add a Rust side function call for `PlatformAudio.DeviceSelectionSupported` for clients to query.

MaxHeimbrock: If we do support device selection on mobile in the future, the client code does not change.

Samples~/Meet/Assets/Runtime/MeetManager.cs
```diff
@@ -62,6 +86,41 @@ private void Start()
...
+            var (recording, playout) = _platformAudio.GetDevices();
+            Debug.Log("Recording devices:");
+            foreach (var device in recording)
+                Debug.Log($"  [{device.Index}] {device.Name}");
+
+            Debug.Log("Playout devices:");
+            foreach (var device in playout)
+                Debug.Log($"  [{device.Index}] {device.Name}");
+
+            if (_platformAudio.RecordingDeviceCount > 0)
```
MaxHeimbrock, line 111: Test with a device that has no input or output device

Runtime/Scripts/PlatformAudio.cs
MaxHeimbrock: Split between input and output device selection?

MaxHeimbrock:
Add tests to init platform audio.

MaxHeimbrock:
> Add tests to init platform audio.

Look at the cpp side:
https://github.com/livekit/client-sdk-cpp/pull/140/changes#diff-bdb98936b19d95cb34bc02aa2a30db943ed77f57b168441ff6fb0b05bca03f70

Runtime/Scripts/PlatformAudio.cs
MaxHeimbrock: Rejected the idea based on @xianshijing-lk feedback

MaxHeimbrock:
> > Add tests to init platform audio.
> 
> Look at the cpp side: https://github.com/livekit/client-sdk-cpp/pull/140/changes#diff-bdb98936b19d95cb34bc02aa2a30db943ed77f57b168441ff6fb0b05bca03f70

Added tests, but on CI there is no platform audio available. Still useful for manual testing on local machines.

Runtime/Plugins/iOS/LiveKitAudioSession.mm
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2024 LiveKit, Inc.
```
xianshijing-lk, line 2: nit, fix the copyright year ?

xianshijing-lk (approved):

PR #301, opened 2026-06-10
Return a PublishDataInstruction from PublishData

MaxHeimbrock:
## Dependency

Depends on https://github.com/livekit/rust-sdks/pull/1137/ 
Fails without: https://github.com/livekit/client-sdk-unity/actions/runs/27275793150

## Summary

`LocalParticipant.PublishData` was fire-and-forget (`void`), so callers had no way to know whether the Rust side accepted the packet — most importantly, packets exceeding the negotiated maximum message size were dropped with only a `log::warn!`.

This wires through the result the FFI transport **already carried** (no proto regen, no Rust changes):
- `PublishDataRequest` already has `request_async_id`; Rust already emits `PublishDataCallback { async_id, error }` and `data_task` forwards the publish result into it. The C# side simply discarded the event.

Changes:
- New `PublishDataInstruction` (`YieldInstruction`) exposing `IsError` + `Error`.
- The three `PublishData` overloads now return it instead of `void` (**source-compatible** — callers ignoring the result still compile).
- Route the callback by adding `PublishData` to `FFIClient.ExtractRequestAsyncId`.

Oversized packets now surface the error via `instruction.Error`:
> data packet size (66602 bytes) exceeds the negotiated maximum message size (64000 bytes)

## Tests

- `ReturnsSuccess_ForSmallPayload` — verifies the success path; runs on any FFI binary.
- `ReturnsError_ForOversizedPayload` — verifies the size-limit error. Marked `[Explicit]`: it requires the `client-sdk-rust` `datamessage_size` FFI binary that enforces the 64000-byte limit (shipped plugins do not yet). Against the old binary an oversized *reliable* send wedges the publisher transport instead of returning cleanly — which is the behavior the limit fixes.

Verified locally: all 2 `PublishDataTests` pass (3.3s) against a macOS FFI lib rebuilt from `datamessage_size` + `livekit-server --dev`.

### DIFF
--- a/Runtime/Scripts/Participant.cs
+++ b/Runtime/Scripts/Participant.cs
@@ -105,18 +105,34 @@ namespace LiveKit
             return instruction;
         }
 
-        public void PublishData(byte[] data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
+        /// <summary>
+        /// Publishes arbitrary data to participants in the room.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="PublishDataInstruction"/> that completes once the packet is sent or errors.
+        /// Check <see cref="YieldInstruction.IsError"/> and read <see cref="PublishDataInstruction.Error"/>
+        /// to handle the result (e.g. payloads exceeding the negotiated maximum message size).
+        /// </returns>
+        public PublishDataInstruction PublishData(byte[] data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
         {
-            PublishData(new Span<byte>(data), destination_identities, reliable, topic);
+            return PublishData(new Span<byte>(data), destination_identities, reliable, topic);
         }
 
-        public void PublishData(Span<byte> data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
+        /// <summary>
+        /// Publishes arbitrary data to participants in the room.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="PublishDataInstruction"/> that completes once the packet is sent or errors.
+        /// Check <see cref="YieldInstruction.IsError"/> and read <see cref="PublishDataInstruction.Error"/>
+        /// to handle the result (e.g. payloads exceeding the negotiated maximum message size).
+        /// </returns>
+        public PublishDataInstruction PublishData(Span<byte> data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
         {
             unsafe
             {
                 fixed (byte* pointer = data)
                 {
-                    PublishData(pointer, data.Length, destination_identities, reliable, topic);
+                    return PublishData(pointer, data.Length, destination_identities, reliable, topic);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


Runtime/Scripts/Participant.cs
```diff
@@ -105,18 +105,34 @@ public UnpublishTrackInstruction UnpublishTrack(ILocalTrack localTrack, bool sto
...
-        public void PublishData(byte[] data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
+        /// <summary>
+        /// Publishes arbitrary data to participants in the room.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="PublishDataInstruction"/> that completes once the packet is sent or errors.
+        /// Check <see cref="YieldInstruction.IsError"/> and read <see cref="PublishDataInstruction.Error"/>
+        /// to handle the result (e.g. payloads exceeding the negotiated maximum message size).
+        /// </returns>
+        public PublishDataInstruction PublishData(byte[] data, IReadOnlyCollection<string> destination_identities = null, bool reliable = true, string topic = null)
```
xianshijing-lk, line 116: Instruction does not seem to be a common technology for such purpose ? 
how about PublishDataTask ?

Runtime/Scripts/Participant.cs
```diff
@@ -661,6 +682,52 @@ internal UnpublishTrackInstruction(ulong asyncId)
...
+
+            if (!string.IsNullOrEmpty(e.Error))
+            {
+                Error = e.Error;
+                IsError = true;
+            }
+            IsDone = true;
+        }
+
+        void OnCanceled()
```
xianshijing-lk, line 723: nit, should onCanceled() private ?

xianshijing-lk (approved):
some nits / questions

Runtime/Scripts/Participant.cs
MaxHeimbrock: We use the `instruction` term since the base type is [CustomYieldInstruction](https://docs.unity3d.com/6000.4/Documentation/ScriptReference/CustomYieldInstruction.html) 

Other examples in our code base:
- [ConnectInstruction](https://github.com/livekit/client-sdk-unity/blob/cd81bf33c86ff37a6684220dfd7223dd54eac137/Runtime/Scripts/Room.cs#L649)
- [PerformRpcInstruction](https://github.com/livekit/client-sdk-unity/blob/cd81bf33c86ff37a6684220dfd7223dd54eac137/Runtime/Scripts/Participant.cs#L670)
- [ReadFrameInstruction](https://github.com/livekit/client-sdk-unity/blob/cd81bf33c86ff37a6684220dfd7223dd54eac137/Runtime/Scripts/DataTrack.cs#L427)

MaxHeimbrock: Instructions use Coroutines underneath, Tasks use C# async

Runtime/Scripts/Participant.cs
MaxHeimbrock: It is private by default, and we also use the implicit private style around in the other instruction classes:
https://github.com/livekit/client-sdk-unity/blob/cd81bf33c86ff37a6684220dfd7223dd54eac137/Runtime/Scripts/Participant.cs#L633

Runtime/Scripts/Participant.cs
xianshijing-lk: I see. Thanks for the context

xianshijing-lk (approved):

YouRMum707 (commented):
👍

PR #304, opened 2026-06-11
Configure native audio source from device config instead of hardcoded defaults

MaxHeimbrock:
## Problem

When the device's audio output configuration differs from the hardcoded defaults — most commonly with a **Bluetooth headset** — microphone capture fails with `sample_rate and num_channels don't match` and the `RtcAudioSource` metadata-mismatch warning.

**Root cause:** `RtcAudioSource` created the native (Rust) audio source with a fixed `sample_rate` (48000) and `num_channels` (2). But captured frames flow through Unity's audio graph (`AudioProbe.OnAudioFilterRead`) at the **actual DSP output configuration**. The Rust native source does not resample — `NativeAudioSource::capture_frame` rejects any frame whose rate/channels differ from how the source was configured.

## Change

- `RtcAudioSource` now reads the sample rate and channel count from Unity's output configuration (`AudioSettings.GetConfiguration`) instead of hardcoded defaults
- The base constructor exposes two overloads: a **device-mode** one (`RtcAudioSourceType` only) for capture sources whose format is whatever Unity delivers, and an **explicit** one (`type, sampleRate, channels`) for sources that generate a fixed, known format, for example `SineWaveAudioSource` in tests

## Verification

Before, using my Sony MDR-1000X Bluetooth headset, I often got 44100 Hz as sample rate from Unity but we configured to 48000 Hz hard coded. This solves the issue for me.

### DIFF
--- a/Runtime/Scripts/BasicAudioSource.cs
+++ b/Runtime/Scripts/BasicAudioSource.cs
@@ -19,9 +19,11 @@ namespace LiveKit
         /// Creates a new basic audio source for the given <see cref="AudioSource"/> in the scene.
         /// </summary>
         /// <param name="source">The <see cref="AudioSource"/> to capture from.</param>
-        /// <param name="channels">The number of channels to capture.</param>
         /// <param name="sourceType">The type of audio source.</param>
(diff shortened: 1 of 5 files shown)


cloudwebrtc (approved):
lgtm!

PR #306, opened 2026-06-12
Reconstruct fragmented mic clip: fragment-aware direct capture

MaxHeimbrock:
Stacked on #304. Fixes the choppy/garbled published audio with a Bluetooth HFP headset mic on macOS — by reading the audio Unity actually delivers, which turned out to be intact but scattered.

## Root cause (proven by buffer inspection)

A raw WAV dump of the mic clip in the bad state showed the exact structure: **FMOD writes each real 20 ms packet of `clip.frequency` audio, then advances `Microphone.GetPosition` as if it had written ~3.2× as much, zero-filling the skipped range.** Concretely: valid fragments of **exactly 320 samples at a stride of exactly 1024** (320/1024 = 1/k where k = counterRate/clip.frequency = 3.2), with exact-zero padding between them (true silence has a real noise floor, never exact zeros).

Junction analysis showed the fragments **join continuously** (boundary sample deltas within normal in-fragment variation) — the full stream is present, just zero-padded. Concatenating fragments reconstructed clean, correct-pitch voice (verified by ear).

This explains every prior symptom:
- plain playback → 31% voice + 69% silence = **chop**
- counter-paced reading → fragments + padding fast over a live buffer = **noise with echo**
- pitch servo (either model) → cannot help; the gaps are in the data layout, not the timing

## Change

`MicrophoneSource` now does **fragment-aware direct capture**:
- Reads the clip ring buffer directly (no `AudioSource`, no `OnAudioFilterRead`) — also decouples capture from the output device's clock.
- A ~0.3 s pre-roll measures `k` (counter rate ÷ `clip.frequency`) and the counter's smallest discrete jump (the stride `J`).
- **k ≈ 1 (healthy devices):** plain contiguous read at the counter's pace.
- **k > 1.05 (this state):** read only the first `J/k` samples of each stride — exactly the valid fragments — skipping the padding.
- Downmix → mono, resample `clip.frequency` → fixed 48 kHz native source (streaming linear; resampler state carries across fragments since junctions are continuous).
- Backlog beyond 200 ms after a stall is dropped (stride-aligned) so the native queue can't overrun.

Expected log in the bad state:
```
MicrophoneSource: fragmented clip detected (k=3.20); reading 320 of every 1024 samples at 16000Hz
```
Healthy devices log `contiguous capture (k=1.00)`.

## Verification

- Runtime compiles clean.
- Buffer-dump analysis (fragment sizes/strides, junction continuity, reconstruction) done offline on a captured WAV; reconstruction validated by ear.
- To validate end-to-end: mac publisher with BT headset mic → Android receiver; expect clean, correct-pitch, non-choppy audio. Built-in mic should behave unchanged.

## History

This branch went through two falsified designs first — a pitch servo at the counter ratio (garbled: the counter doesn't describe the data) and a k-rescaled lag servo at pitch 1 (perfect telemetry, still choppy: the gaps are in the buffer itself). The WAV dump diagnostic settled it. Commits preserved for the record.

### DIFF
--- a/Runtime/Scripts/AudioStream.cs
+++ b/Runtime/Scripts/AudioStream.cs
@@ -50,6 +50,14 @@ namespace LiveKit
         private const int CrossfadeFrames = 128;  // ~2.7ms @ 48kHz
         private int _skipCooldown = 0;
 
+        // --- Temporary receive diagnostics (Info level, emitted ~every 2s) ---
+        // Reveals whether choppiness is a buffer-starvation problem (underruns/low fill) versus a
+        // clean stream, and what rate/channels we are actually playing/requesting.
+        private long _diagWindowStartTicks;
+        private int _diagCallbacks;
+        private int _diagUnderruns;
+        private int _diagFramesReceived;
+
         /// <summary>
         /// Creates a new audio stream from a remote audio track, attaching it to the
         /// given <see cref="AudioSource"/> in the scene.
@@ -147,6 +155,8 @@ namespace LiveKit
 
             lock (_lock)
             {
+                MaybeLogReceiveDiagnostics(channels, sampleRate);
+
                 // Single gate covering first-create and runtime format changes (e.g. after a
                 // system audio device switch). When the FFI stream is missing or what we asked
                 // Rust for no longer matches what Unity is delivering, post a (re)create to the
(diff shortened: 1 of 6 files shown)


MaxHeimbrock:
MDR-1000X microphone is fixed in Unity 6, so no need for this.

PR #312, opened 2026-06-15
Recreate audio source and republish track on sample-rate change

MaxHeimbrock:
> **Stacked on #311** — review/merge that first. This PR's diff is one commit on top of it.

## Problem

#311 restarts capture on a device change, but the native audio source's rate is fixed at construction. If the new device has a different sample rate, `RtcAudioSource` silently drops every captured frame (it rejects frames whose rate/channels don't match the source) — so audio stays dead.

The FFI exposes no in-place source reconfigure (only `NewAudioSource`), so recovery requires a fresh native source and a re-bound track.

## Change

- Add `RtcAudioSource.Reconfigure(sampleRate, channels)`: disposes the old native handle, rebuilds the source at the new format, and raises a new `FormatChanged` event. The native source stays alive via the track's reference until the track is dropped, so disposing the handle before the old track is unpublished is safe.
- `MicrophoneSource` detects the format change via `ResolveDeviceFormat` in its config-changed handler and calls `Reconfigure` **inside the restart while capture is paused** (no `AudioRead` callbacks in flight), replacing the warning from #311.
- `MeetManager` subscribes to `FormatChanged` and republishes the local audio track against the source's new handle (unsubscribing on unpublish/cleanup).

## Note

Republishing yields a new track SID, so remote peers resubscribe on a rate change — acceptable but observable.

## Testing

- `Scripts~/run_unity.sh build macos` → Build SUCCEEDED.
- Manual runtime verification (device swap that actually changes the DSP rate → audio recovers with a republish log line) still recommended on a real device.

### DIFF
--- a/Runtime/Scripts/RtcAudioSource.cs
+++ b/Runtime/Scripts/RtcAudioSource.cs
@@ -83,20 +83,33 @@ namespace LiveKit
         private volatile bool _disposed = false;
         private int _audioReadCount = 0;
 
-        protected RtcAudioSource(int channels = 2, RtcAudioSourceType audioSourceType = RtcAudioSourceType.AudioSourceCustom)
+        // Device-capture sources (microphone, AudioSource taps) don't know their format ahead of
+        // time — it is whatever Unity's audio graph delivers. They use this constructor, which
+        // configures the native source from Unity's current output configuration.
+        protected RtcAudioSource(RtcAudioSourceType audioSourceType)
+            : this(audioSourceType, 0, 0) { }
+
+        // Sources that generate a fixed, known format (e.g. test signal generators) declare it
+        // directly. Passing 0 for either value falls back to the device configuration.
+        protected RtcAudioSource(RtcAudioSourceType audioSourceType, uint sampleRate, uint channels)
         {
             _sourceType = audioSourceType;
-            _expectedChannels = (uint)channels;
+
+            if (sampleRate > 0 && channels > 0)
+            {
+                _expectedSampleRate = sampleRate;
+                _expectedChannels = channels;
+            }
+            else
+            {
+                (_expectedSampleRate, _expectedChannels) = ResolveDeviceFormat();
+            }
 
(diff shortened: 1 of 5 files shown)


Runtime/Scripts/RtcAudioSource.cs
```diff
@@ -111,11 +129,41 @@ protected RtcAudioSource(RtcAudioSourceType audioSourceType, uint sampleRate, ui
...
+        /// on success so the owner can re-bind any track to the new handle.
+        /// </remarks>
+        /// <returns>True if the source was recreated; false if the format was unchanged or invalid.</returns>
+        public bool Reconfigure(uint sampleRate, uint channels)
+        {
+            if (_disposed) return false;
+            if (sampleRate == 0 || channels == 0) return false;
+            if (sampleRate == _expectedSampleRate && channels == _expectedChannels) return false;
+
+            Utils.Debug($"{DebugTag} reconfigure {_expectedSampleRate}/{_expectedChannels} -> {sampleRate}/{channels}");
```
MaxHeimbrock, line 149: Without this change, we can have:
```
LiveKit: MicrophoneSource: audio device change moved the DSP output rate to 44100Hz, but the native source is fixed at 48000Hz. Captured frames will be dropped until the track is recreated at the new rate.
UnityEngine.Logger:LogWarning (string,object)
LiveKit.Internal.Utils:Warning (object) (at /Users/maxheimbrock/dev/unity/client-sdk-unity/Runtime/Scripts/Internal/Utils.cs:41)
LiveKit.MicrophoneSource:OnAudioConfigurationChanged (bool) (at /Users/maxheimbrock/dev/unity/client-sdk-unity/Runtime/Scripts/MicrophoneSource.cs:248)
UnityEngine.AudioSettings:InvokeOnAudioConfigurationChanged (bool) (at /Users/bokken/build/output/unity/unity/Modules/Audio/Public/ScriptBindings/Audio.bindings.cs:413)
```

PR #319, opened 2026-06-17
Complete UniTask implementation

MaxHeimbrock:
### Background

We currently only support Coroutines for async execution, with this PR we will support async/await and if clients import the UniTask package, we also support UniTasks natively.


Runtime/Scripts/UniTask/StreamReaderUniTaskExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+            return UniTaskAsyncEnumerable.Create<TChunk>(async (writer, token) =>
+            {
+                // The enumerator hands us its own token; honor both it and the caller's.
+                using var linked = CancellationTokenSource.CreateLinkedTokenSource(token, cancellationToken);
+                var ct = linked.Token;
+
+                while (true)
+                {
+                    // Completes when a chunk is ready (IsCurrentReadDone) or the stream ends (IsEos).
+                    await instruction.AsUniTask(ct);
```
xianshijing-lk, line 48: what will happen if cancellation is requested before or after "await instruction.AsUniTask(ct);" ?

Runtime/Scripts/UniTask/YieldInstructionUniTaskExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// running and any result is discarded. Wire-level cancellation is not yet
+        /// supported. Error inspection stays on the instruction itself — the awaiter does
+        /// not throw on <see cref="YieldInstruction.IsError"/>, matching the existing
+        /// <c>yield return</c> / <c>await</c> behavior.
+        /// </remarks>
+        public static UniTask AsUniTask(this YieldInstruction instruction, CancellationToken cancellationToken = default)
+        {
+            if (instruction == null) throw new System.ArgumentNullException(nameof(instruction));
+            if (instruction.IsDone) return UniTask.CompletedTask;
+            if (cancellationToken.IsCancellationRequested) return UniTask.FromCanceled(cancellationToken);
```
xianshijing-lk, line 32: what will happen if calling AsUnitTask twice on the same instruction ? should we protect it ?

xianshijing-lk (approved):
two questions / nits. lgtm

Runtime/Scripts/UniTask/YieldInstructionUniTaskExtensions.cs
MaxHeimbrock: I tested it right now and it does result in undefined behaviour, but I don't think we should protect it. I think there are many ways the client code can make mistakes here, we just assume they don't.

Runtime/Scripts/UniTask/StreamReaderUniTaskExtensions.cs
MaxHeimbrock: When I cancel the token before awaiting, the instruction is cancelled immediately after start, as expected:
```
var cts = new CancellationTokenSource();
var token = cts.Token;
cts.Cancel();

var connect = _room.Connect(details.ServerUrl, details.ParticipantToken, new RoomOptions());
await connect.AsUniTask(token);
```
=> `OperationCanceledException: The operation was canceled.`

Cancelling after awaiting has no effect, also as expected.

PR #335, opened 2026-07-01
Add the agents sample to the package

MaxHeimbrock:
### Background

Adds a new sample that lets you connect to a LiveKit agent and get the transcription in form of a chat log.
Per default, the "Haley" agent from our website is configured as a TokenSource.

Also the ScrollingLog we use in both samples is pulled out into a Common sample, which needs to be imported in order for the ScrollingLog script to be loaded in the Meet and Agents sample.


Samples~/Agents/Assets/Runtime/Agent/LiveKitAgentSession.cs
```diff
@@ -0,0 +1,133 @@
+using System;
+using System.Collections;
+using LiveKit;
+using UnityEngine;
+using RoomOptions = LiveKit.RoomOptions;
+
+namespace AgentsRPG
```
MaxHeimbrock: Realized the old namespace is still here.

MaxHeimbrock: MeetManager.cs has no namespace, so I will also remove this one.

MaxHeimbrock: Done

ladvoc (approved):

ladvoc:
question: Should MP3s and JPG be added to Git LFS?

MaxHeimbrock:
> question: Should MP3s and JPG be added to Git LFS?

Good idea! Done

PR #350, opened 2026-07-13
Restore cached iOS audio session on PlatformAudio dispose

MaxHeimbrock:
## Problem

On iOS, constructing `PlatformAudio` calls `LiveKit_ConfigureAudioSessionForVoIP()`, switching the shared `AVAudioSession` to `PlayAndRecord` + `VoiceChat`. On dispose, nothing restored it:

- `PlatformAudio.Dispose()` never called the restore hook — `LiveKit_RestoreDefaultAudioSession()` was **dead code** (declared + implemented, zero callers).
- Even if called, it hardcoded `AVAudioSessionCategoryAmbient` and never reactivated.

Net effect: after a call ended, the app's session stayed on the VoIP profile and deactivated, so normal Unity audio playback was degraded (quiet, receiver-biased routing). `AudioSettings.Reset` does not fix this — it only reinitializes Unity's internal (FMOD) engine, not the OS-level session.

## Change

- **`LiveKitAudioSession.mm`**: snapshot `category`/`mode`/`categoryOptions` once, lazily, at the top of `LiveKit_ConfigureAudioSessionForVoIP()` — i.e. right before LiveKit first mutates the session, so it captures whatever Unity configured from its iOS Player Settings. Rewrite `LiveKit_RestoreDefaultAudioSession()` to re-apply that snapshot and `setActive:YES` (falls back to `Ambient` only if nothing was cached).
- **`PlatformAudio.cs`**: track live instances with an `Interlocked` counter (iOS-guarded); increment after successful construction, decrement in `Dispose()`, and fire restore only when the count hits 0 — matching the native ADM ref-count teardown.

Explicit-`Dispose()` only (no finalizer): avoids touching `AVAudioSession` off the main thread. Kept minimal — the existing automatic-mode / `VoiceChat` config is unchanged.

## Testing

- C# Runtime assembly compiles cleanly with `UNITY_IOS` defined / `UNITY_EDITOR` off (0 errors) — the guarded code is actually type-checked.
- `.mm` passes `clang -fsyntax-only` against the iOS SDK under both ARC and MRC (no errors).
- **Needs on-device verification** (not doable in CI): connect a room with `PlatformAudio`, dispose it, then play a Unity `AudioSource` and confirm routing/volume return to the pre-call behavior. The new `NSLog`s report the cached + restored category/mode.

## Follow-ups (out of scope)

- A more advanced manual-mode prototype (RTCAudioSession `useManualAudio`, `VideoChat`, VPIO gating to keep Unity audio alive *during* calls) exists as a build artifact under `Samples~/Meet/Builds/` — not adopted here.
- `.mm.meta` also targets visionOS/tvOS, but the C# P/Invoke is `UNITY_IOS`-only (pre-existing).

### DIFF
--- a/Runtime/Plugins/iOS/LiveKitAudioSession.mm
+++ b/Runtime/Plugins/iOS/LiveKitAudioSession.mm
@@ -16,6 +16,32 @@
 
 #import <AVFoundation/AVFoundation.h>
 
+// Snapshot of the AVAudioSession configuration as it was the first time LiveKit
+// touched the session (i.e. whatever Unity set up from its iOS Player Settings).
+// Captured lazily in LiveKit_ConfigureAudioSessionForVoIP and re-applied by
+// LiveKit_RestoreDefaultAudioSession when the last PlatformAudio is disposed.
+static BOOL s_hasCachedState = NO;
+static NSString* s_cachedCategory = nil;
+static NSString* s_cachedMode = nil;
+static AVAudioSessionCategoryOptions s_cachedCategoryOptions = 0;
+
+/// Captures the current audio session category/mode/options exactly once, before
+/// LiveKit reconfigures the session for VoIP. Subsequent calls are no-ops so the
+/// snapshot always reflects the pristine, pre-LiveKit (Unity-configured) state.
+static void LiveKit_CacheSessionStateIfNeeded() {
+    if (s_hasCachedState) {
+        return;
+    }
+    AVAudioSession* session = [AVAudioSession sharedInstance];
+    // copy so the strings persist for the app lifetime regardless of ARC/MRC.
+    s_cachedCategory = [session.category copy];
+    s_cachedMode = [session.mode copy];
+    s_cachedCategoryOptions = session.categoryOptions;
+    s_hasCachedState = YES;
+    NSLog(@"LiveKit: cached audio session state: category=%@, mode=%@, options=%lu",
+          s_cachedCategory, s_cachedMode, (unsigned long)s_cachedCategoryOptions);
+}
+
 extern "C" {
 
 /// Configures the iOS audio session for VoIP/WebRTC use.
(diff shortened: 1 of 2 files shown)


Runtime/Plugins/iOS/LiveKitAudioSession.mm
```diff
@@ -16,6 +16,32 @@
...
+static BOOL s_hasCachedState = NO;
+static NSString* s_cachedCategory = nil;
+static NSString* s_cachedMode = nil;
+static AVAudioSessionCategoryOptions s_cachedCategoryOptions = 0;
+
+/// Captures the current audio session category/mode/options exactly once, before
+/// LiveKit reconfigures the session for VoIP. Subsequent calls are no-ops so the
+/// snapshot always reflects the pristine, pre-LiveKit (Unity-configured) state.
+static void LiveKit_CacheSessionStateIfNeeded() {
+    if (s_hasCachedState) {
```
xianshijing-lk, line 32: curiously, should we continue caching the latest session rather than simply return here ?

xianshijing-lk (approved):
one question, lgtm

Runtime/Scripts/Audio/PlatformAudio.cs
```diff
@@ -365,6 +376,14 @@ public void Dispose()
             if (_disposed) return;
             Handle.Dispose();
             _disposed = true;
+
+#if UNITY_IOS && !UNITY_EDITOR
+            // Restore the audio session Unity had before VoIP, but only once the
+            // last PlatformAudio is gone (the native ADM is likewise ref-counted).
+            if (System.Threading.Interlocked.Decrement(ref _instanceCount) == 0)
+                IOSAudioSessionHelper.LiveKit_RestoreDefaultAudioSession();
```
xianshijing-lk, line 384: nit, add a {} for style

Runtime/Scripts/Audio/PlatformAudio.cs
```diff
@@ -74,6 +74,11 @@ public sealed class PlatformAudio : IDisposable
         internal readonly FfiHandle Handle;
         private readonly PlatformAudioInfo _info;
         private bool _disposed = false;
+#if UNITY_IOS && !UNITY_EDITOR
+        // Tracks live PlatformAudio instances so the iOS audio session is restored
+        // only when the last one is disposed (aligned with the native ADM ref-count).
+        private static int _instanceCount;
```
xianshijing-lk, line 80: nit, initialize it to 0 as default value ?

Runtime/Plugins/iOS/LiveKitAudioSession.mm
MaxHeimbrock: Sorry this is supposed to be a draft, not ready for review. We have some issues with teardown but this one does not solve it all, I am still investigating.

PR #354, opened 2026-07-15
Fail loudly on file not found

MaxHeimbrock:
### Background

When we send a file and the file does not exist in the path, fail.


MaxHeimbrock:
The right pattern is to check on the error of the returned instruction. This is not needed and I will update the docs.

PR #357, opened 2026-07-20
Fix bug in EOS for data streams

MaxHeimbrock:
### Background

When a burst of packages plus the EOS trailer arrived at the same time, the incremental reading of the stream was broken. If the stream was marked EOS, but the queue still had chunks, they were ignored.

This PR adds the logic to fix the bug plus tests.  

### Changes

What did you do?

- Fix bug when EOS message could mark the stream as ended, even when queue was non-empty
- Add tests to guard the expected behaviour
- Simplify agent sample code since workaround is not needed anymore

### DIFF
--- a/Runtime/Scripts/DataStreams/DataStream.cs
+++ b/Runtime/Scripts/DataStreams/DataStream.cs
@@ -85,6 +85,8 @@ namespace LiveKit
 
         // A chunk or the terminal end-of-stream marker. Chunks and the marker share one
         // FIFO so the consumer can never observe end-of-stream while chunks remain buffered.
+        // The head of the queue is the item the consumer is currently positioned on; Reset()
+        // dequeues it to advance to the next.
         private readonly struct Item
         {
             public readonly TContent Chunk;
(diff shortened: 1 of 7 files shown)


Tests/EditMode/DataStreamIncrementalReadTests.cs
```diff
@@ -109,20 +108,128 @@ public void OnEos_RacingWithChunks_FinalStateConsistent()
...
+        {
+            using var handle = new FfiHandle(IntPtr.Zero);
+            var reader = new TestIncrementalReader(handle);
+
+            reader.PushChunk("A");
+            reader.PushChunk("B");
+            reader.PushChunk("C");
+            reader.PushEos();
+
+            CollectionAssert.AreEqual(new[] { "A", "B", "C" }, Drain(reader));
```
MaxHeimbrock, line 155: If all chunks arrived in the same frame together with EOS, before the fix the incremental reader would receive EOS and just be marked as done, disregarding the chunks in the queue.

Runtime/Scripts/DataStreams/DataStream.cs
```diff
@@ -72,19 +72,43 @@ internal StreamError(Proto.StreamError proto) : base(proto.Description) { }
...
-        // run on the main-thread coroutine. _gate serializes mutations of the queue,
-        // _latestChunk, IsCurrentReadDone, IsEos, and Error across both sides.
+        // A chunk or the terminal end-of-stream marker. Chunks and the marker share one
+        // FIFO so the consumer can never observe end-of-stream while chunks remain buffered.
+        // The head of the queue is the item the consumer is currently positioned on; Reset()
+        // dequeues it to advance to the next.
+        private readonly struct Item
+        {
+            public readonly TContent Chunk;
+            public readonly bool IsTerminal;
```
xianshijing-lk: nitpick, why the variable is called "isTerminal" ? I find it a bit hard to understand what it does from the name.

MaxHeimbrock: I aligned it to also be called IsEos to not add more keywords.

xianshijing-lk (approved):
lgtm

PR #360, opened 2026-07-24
Fix iOS build on Unity 6.5

### LINKED ISSUE #359: iOS Unity 6 crash in ec_tell_frac: IosLinkOrderDiagnostics does not handle libGameAssembly.a
## Summary

Our Unity iOS application crashes only on iOS when LiveKit audio is initialized or when a remote audio track is subscribed.

Android builds using the same LiveKit integration work normally.

The crash consistently stops in `UnityFramework` at the Opus/CELT symbol `ec_tell_frac`, suggesting a native codec object/symbol conflict.

## Environment

- LiveKit Unity SDK: `2.0.0`
- Unity: `6000.5.5f1`
- Scripting backend: IL2CPP
- Architecture: arm64
- Target: physical iOS device
- iOS deployment target: 15.0
- Installation: UPM Git URL (`https://github.com/livekit/client-sdk-unity.git`)

## Symptoms

We observed crashes at multiple audio lifecycle points:

1. Immediately after local microphone initialization:

```text
LiveKit: Configured native audio source with sampleRate 48000 and channels 2
LiveKit.RtcAudioSource:ResolveDeviceFormat()
LiveKit.RtcAudioSource:.ctor(...)
LiveKit.MicrophoneSource:.ctor(...)

2. After remote audio subscription:
[LiveKitVoiceChatMgr] Remote audio subscribed (AudioStream playout): <participant>

3. The debugger then stops in:
UnityFramework`ec_tell_frac
In some attempts, the process instead crashes around:

libsystem_pthread.dylib`___chkstk_darwin
Android works with the same server, token flow, room configuration, and application code.

Relevant SDK behavior
LiveKit SDK 2.0.0 includes Editor/IosLinkOrderDiagnostics.cs.

This file documents that linking Unity's audio archive before liblivekit_ffi.a can crash Opus/CELT on iOS. It also strips conflicting CELT object files from:

Libraries/libiPhone-lib.a
However, Unity 6 exports libGameAssembly.a rather than libiPhone-lib.a.

Therefore, the current post-process logic does not appear to process the Unity 6 archive. In our build logs, we do not see the expected CELT stripping log, and the crash still occurs in ec_tell_frac.

Expected behavior
The SDK’s iOS post-process diagnostics/fix should also support Unity 6 exports using libGameAssembly.a, including:

Safe link ordering of liblivekit_ffi.a before the Unity-generated archive.
Detection and removal of conflicting CELT/Opus object files from the applicable Unity-generated archive.
Clear build-log output stating which archive was detected and modified.

Actual behavior
The current fix targets libiPhone-lib.a, which is absent or unused in Unity 6 exports. The resulting iOS application crashes in the CELT/Opus path while Android remains unaffected.

Question
Could the iOS link-order/CELT conflict mitigation be updated to support libGameAssembly.a for Unity 6 / IL2CPP exports? If there is a preferred integration path or a known fix for this configuration, please advise.

### PR DESCRIPTION
**Pull request by MaxHeimbrock**
### Background

This fixes https://github.com/livekit/client-sdk-unity/issues/359

Unity 6000.5 re-architected the iOS export: the engine library is no longer
`Libraries/libiPhone-lib.a` but a static `Frameworks/UnityRuntime.framework`
(which still bundles the old CELT/libvpx objects that conflict with the opus/vpx
inside `liblivekit_ffi.a`). Our iOS post-build protection was hardwired to
`libiPhone-lib.a`, so on 6000.5 it silently no-oped and apps crashed at runtime
in `ec_tell_frac` (confirmed on device with the Meet sample).

On top of that, the Meet and Agents samples' own `BuildPostProcessor` passed the
now-null file GUID for `libiPhone-lib.a` into `RemoveFileFromBuild`, failing every
iOS build on 6000.5 with an `ArgumentNullException`.

The exported frameworks build phase is also no longer labeled `/* UnityFramework */`;
build logs show that marker didn't even match on 6000.3, so the link-order rewrite
was already a no-op there and only the CELT strip was protecting older builds.

### Changes

- `IosLinkOrderDiagnostics` now resolves the engine library per export layout:
  `Libraries/libiPhone-lib.a` (up to 6000.4) or `Frameworks/UnityRuntime.framework`
  (6000.5+). Reorder, CELT strip, and the diagnostic all operate on the resolved
  library; behavior on the old layout is unchanged.
- Locate the UnityFramework frameworks build phase by content (the phase linking
  `liblivekit_ffi.a`) instead of the `/* UnityFramework */` label, which also
  repairs the reorder/diagnostic on pre-6.5 exports.
- Meet & Agents samples: guard the `libiPhone-lib.a` remove/re-add so iOS builds
  no longer fail on 6000.5.
- README: document both export layouts.

Verified on Unity 6000.5.5f1 (device): export logs show the link-order fix applied,
18 conflicting CELT objects stripped from `UnityRuntime`, diagnostic OK, and the
Meet app no longer crashes.

### DIFF
--- a/Editor/IosLinkOrderDiagnostics.cs
+++ b/Editor/IosLinkOrderDiagnostics.cs
@@ -13,7 +13,7 @@ public static class IosLinkOrderDiagnostics
 {
     private const string Prefix = "LiveKit";
     private const string Libilivekit = "liblivekit_ffi.a in Frameworks";
-    // Unity exports the engine as Libraries/libiPhone-lib.a up to 6000.3 and as a static
+    // Unity exports the engine as Libraries/libiPhone-lib.a up to 6000.4 and as a static
     // Frameworks/UnityRuntime.framework from 6000.5; both bundle the old CELT objects.
     private static readonly string[] EngineLibraryEntries =
     {
--- a/Editor/IosLinkOrderDiagnostics.cs
+++ b/Editor/IosLinkOrderDiagnostics.cs
@@ -13,8 +13,14 @@ public static class IosLinkOrderDiagnostics
 {
     private const string Prefix = "LiveKit";
     private const string Libilivekit = "liblivekit_ffi.a in Frameworks";
-    private const string Libiphone = "libiPhone-lib.a in Frameworks";
-    private static readonly string[] ConflictingLibiPhoneMembers =
+    // Unity exports the engine as Libraries/libiPhone-lib.a up to 6000.3 and as a static
+    // Frameworks/UnityRuntime.framework from 6000.5; both bundle the old CELT objects.
+    private static readonly string[] EngineLibraryEntries =
+    {
+        "libiPhone-lib.a in Frameworks",
+        "UnityRuntime.framework in Frameworks"
+    };
+    private static readonly string[] ConflictingEngineArchiveMembers =
     {
         "bands.o",
         "celt.o",
@@ -50,12 +56,14 @@ public static class IosLinkOrderDiagnostics
         }
 
         var projectText = File.ReadAllText(projectPath);
-        var fixedProjectText = EnsureSafeUnityFrameworkLinkOrder(projectText, out var wasModified);
+        var engineEntry = EngineLibraryEntries.FirstOrDefault(entry => projectText.Contains(entry));
+        var engineName = engineEntry?.Replace(" in Frameworks", string.Empty);
+        var fixedProjectText = EnsureSafeUnityFrameworkLinkOrder(projectText, engineEntry, out var wasModified);
         if (wasModified)
         {
             File.WriteAllText(projectPath, fixedProjectText);
             projectText = fixedProjectText;
-            Debug.Log($"{Prefix}: iOS link-order fix applied. Moved libiPhone-lib.a after liblivekit_ffi.a in UnityFramework -> Frameworks and Libraries.");
+            Debug.Log($"{Prefix}: iOS link-order fix applied. Moved {engineName} after liblivekit_ffi.a in UnityFramework -> Frameworks and Libraries.");
         }
 
         StripConflictingCodecObjects(pathToBuiltProject);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Editor/IosLinkOrderDiagnostics.cs [hunk 1]
```diff
@@ -68,64 +76,67 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToBuiltProj
...
         {
-            Debug.Log($"{Prefix}: iOS link-order diagnostic OK. UnityFramework links liblivekit_ffi.a before libiPhone-lib.a.");
+            Debug.Log($"{Prefix}: iOS link-order diagnostic OK. UnityFramework links liblivekit_ffi.a before {engineName}.");
             return;
         }
 
         Debug.LogWarning(
             $"{Prefix}: iOS link-order diagnostic found a risky order in UnityFramework -> Frameworks and Libraries. " +
-            "libiPhone-lib.a appears before liblivekit_ffi.a. This repo documents that this can crash Opus/CELT on iOS. " +
+            $"{engineName} appears before liblivekit_ffi.a. This repo documents that this can crash Opus/CELT on iOS. " +
```
xianshijing-lk, line 94: Don't we have script to auto fix the problem and make sure we load liblivekit_ffi.a before libiPhone-lib.a  ?

xianshijing-lk (commented):
lgtm in general, one question

xianshijing-lk (approved):
lgtm, I just missed the BuildPostProcessor.cs. Please ignore my previous comment

Editor/IosLinkOrderDiagnostics.cs
MaxHeimbrock: Yes, the script works until Unity 6.4, but in 6.5 Unity apparently changed their lib composition and naming. 

A customer reported the crash: https://github.com/livekit/client-sdk-unity/issues/359
They also already confirmed the fix works for them as well. I will now test if it still works in Unity 2022 or Unity 6.3, but that behaviour should still be intact.

MaxHeimbrock: Confirmed the old way of building is not touched in lower Unity versions.

PR #366, opened 2026-07-31
Deprecate sandbox token server and use development token server

MaxHeimbrock:
### Background

Deprecating "Sandbox" token source and renaming it to development token source. 

### Changes

- Obsolete `TokenSourceSandbox` everywhere
- Alias new `TokenSourceDevelopment`
- Update docs

Newer changes:
- Align with JS SDK to use static factory pattern

### DIFF
--- a/Runtime/Scripts/TokenSource/TokenSource.cs
+++ b/Runtime/Scripts/TokenSource/TokenSource.cs
@@ -9,6 +9,197 @@ using Newtonsoft.Json;
 namespace LiveKit
 {
     /// <summary>
+    /// Factory for the built-in <see cref="ITokenSource"/> implementations. The concrete implementations
+    /// are private; obtain them through the factory methods and work with the returned
+    /// <see cref="ITokenSourceFixed"/> or <see cref="ITokenSourceConfigurable"/>.
+    /// </summary>
+    public static class TokenSource
+    {
+        public delegate Task<ConnectionDetails> CustomTokenFunction();
+
+        /// <summary>
+        /// Returns a fixed server URL and participant token. Suitable when credentials are pregenerated
+        /// (e.g. via the LiveKit CLI or LiveKit Cloud project page).
+        /// </summary>
+        public static ITokenSourceFixed Literal(string serverUrl, string participantToken)
+        {
+            return new LiteralSource(serverUrl, participantToken);
+        }
+
+        /// <summary>
+        /// Posts a JSON request to a token-server endpoint and returns the parsed <see cref="ConnectionDetails"/>.
+        /// The body is built from per-call <see cref="TokenSourceFetchOptions"/> (room name, participant info,
+        /// agent dispatch, etc.). Use for production token servers — see
+        /// https://docs.livekit.io/frontends/build/authentication/endpoint/.
+        /// </summary>
+        public static ITokenSourceConfigurable Endpoint(string endpointUrl, IEnumerable<StringPair> headers)
+        {
+            return new EndpointSource(endpointUrl, headers);
+        }
+
+        /// <summary>
+        /// Delegates connection-detail retrieval to a user-supplied async function. Use this when your
+        /// app already has its own token-fetching code (custom auth flow, cached tokens, etc.).
+        /// </summary>
+        public static ITokenSourceFixed Custom(CustomTokenFunction customTokenFunction)
+        {
+            return new CustomSource(customTokenFunction);
+        }
+
+        [Obsolete("Use TokenSource.DevelopmentTokenServer instead")]
+        public static ITokenSourceConfigurable SandboxTokenServer(string sandboxId)
+        {
+            return DevelopmentTokenServer(sandboxId);
+        }
+
+        /// <summary>
+        /// Convenience <see cref="Endpoint"/> preconfigured for LiveKit Cloud development token servers.
+        /// Intended for development and testing only — see
+        /// https://docs.livekit.io/frontends/build/authentication/sandbox-token-server/.
+        /// </summary>
+        public static ITokenSourceConfigurable DevelopmentTokenServer(string tokenServerId)
+        {
+            return new EndpointSource(
+                "https://cloud-api.livekit.io/api/v2/sandbox/connection-details",
+                new[] { new StringPair { key = "X-Sandbox-ID", value = tokenServerId } });
+        }
+
+        private sealed class LiteralSource : ITokenSourceFixed
+        {
+            private string _serverUrl;
+            private string _participantToken;
+
+            public LiteralSource(string serverUrl, string participantToken)
+            {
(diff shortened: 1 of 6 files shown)


Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -34,163 +227,55 @@ public interface ITokenSourceConfigurable : ITokenSource
         public TaskYieldInstruction<ConnectionDetails> FetchConnectionDetails(TokenSourceFetchOptions options);
     }
 
-    /// <summary>
-    /// Returns a fixed server URL and participant token. Suitable when credentials are pregenerated
-    /// (e.g. via the LiveKit CLI or LiveKit Cloud project page).
-    /// </summary>
+    [Obsolete("Use TokenSource.Literal(...) instead")]
```
lukasIO, line 231: question: will the deprecation notice only show up when constructing it or also when using an explicit type like 

```c#
TokenSourceLiteral myTokenSource = TokenSource.Literal(...)
```

MaxHeimbrock: Only when using these specific public constructors.

lukasIO (approved):

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
+                    task = Task.FromException<ConnectionDetails>(e);
+                }
+                return new TaskYieldInstruction<ConnectionDetails>(task);
+            }
+        }
+
+        private sealed class TokenSourceEndpoint : ITokenSourceConfigurable
+        {
+            private string _endpointUrl;
+            IEnumerable<StringPair> _headers;
```
xianshijing-lk: nit, should it be private readonly IReadOnlyList<StringPair> _headers; ?

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
+            private static readonly HttpClient HttpClient = new HttpClient();
+
+            public TokenSourceEndpoint(string endpointUrl, IEnumerable<StringPair> headers)
+            {
+                _endpointUrl = endpointUrl;
+                _headers = headers;
```
xianshijing-lk: I wonder if we should just copy the _headers like 
_headers = headers?.ToList() ?? Array.Empty<StringPair>();

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
+        {
+            return new TokenSourceEndpoint(
+                DevelopmentTokenServerUrl,
+                new[] { new StringPair { key = "X-Sandbox-ID", value = tokenServerId } });
+        }
+
+        private sealed class TokenSourceLiteral : ITokenSourceFixed
+        {
+            private string _serverUrl;
+            private string _participantToken;
```
xianshijing-lk: nit, should these fields be readonly ?

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
 namespace LiveKit
 {
+    /// <summary>
+    /// Factory for the built-in <see cref="ITokenSource"/> implementations. The concrete implementations
+    /// are private; obtain them through the factory methods and work with the returned
+    /// <see cref="ITokenSourceFixed"/> or <see cref="ITokenSourceConfigurable"/>.
+    /// </summary>
+    public static class TokenSource
+    {
+        public delegate Task<ConnectionDetails> CustomTokenFunction();
```
xianshijing-lk, line 18: should we make such CustomTokenFunction cancellable ?

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
+                // request lives in the helper below; the returned task carries any synchronous throw.
+                return new TaskYieldInstruction<ConnectionDetails>(FetchConnectionDetailsAsync(options));
+            }
+
+            private async Task<ConnectionDetails> FetchConnectionDetailsAsync(TokenSourceFetchOptions options)
+            {
+                var requestBody = BuildRequest(options);
+                var jsonBody = JsonConvert.SerializeObject(requestBody);
+
+                var request = new HttpRequestMessage(HttpMethod.Post, _endpointUrl);
```
xianshijing-lk: should it be 
using var request = new HttpRequestMessage(...);
so that it could avoid unnecessary allocations/resources ?

Runtime/Scripts/TokenSource/TokenSource.cs
```diff
@@ -8,6 +8,199 @@
...
+                    {
+                        if (!string.IsNullOrEmpty(header.key))
+                            request.Headers.TryAddWithoutValidation(header.key, header.value);
+                    }
+                }
+                var content = new StringContent(jsonBody, System.Text.Encoding.UTF8);
+                content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/json");
+                request.Content = content;
+
+                var response = await HttpClient.SendAsync(request);
```
xianshijing-lk: ditto, does it make sense to be 
using var response = await HttpClient.SendAsync(request);

xianshijing-lk (approved):
some questions, lgtm assuming you will be addressing them

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: That would need a lot more work and in JS we also don't offer that. Maybe something to keep in mind for the future or when customers request it.

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks, done

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks, done

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks, done

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks, done

Runtime/Scripts/TokenSource/TokenSource.cs
MaxHeimbrock: Thanks, done

PR #378, opened 2026-08-26
PHASE A: Add device routing and better session management to C# Unity Platform Audio

MaxHeimbrock:
### Background

We had several issues with platform audio on mobile (iOS & Android), especially with:
- App backgrounding
- Unity audio playing parallel to WebRTC audio
- Device switching during calls (Bluetooth to speaker and back)
- Device selection

### Changes

What did you do?

- Device selection via priority list
- Improved session management not messing or getting messed up by with Unity's audio engine
- Events and watchdog (there are some edge cases where events don't fire correctly) to react to device switches
- The platform's call audio session follows `Room` connections automatically — no client-side call-boundary API
- Includes #382 (`Room.Disconnected` reported for local disconnects too) by merge; it shows up in this diff until #382 lands on `main`

### Public API changes

Everything below is in namespace `LiveKit`, in `Runtime/Scripts/Audio/PlatformAudio.cs` unless noted. The new `IRouteController` / `DesktopRouteController` / `UnsupportedRouteController` / `AndroidRouteController` / `IosRouteController` types are all `internal`.

#### Added

| Member | Signature / notes |
|---|---|
| `AudioDeviceKind` | new `public enum` — `Unknown=0, Earpiece=1, Speaker=2, WiredHeadset=3, Bluetooth=4, Usb=5, HearingAid=6` (numbering mirrors the planned FFI enum of the same name, do not renumber) |
| `AudioDevice.Kind` | `public AudioDeviceKind Kind` — classified for playout devices on iOS and Android 12+; `Unknown` elsewhere |
| `AudioDevice.IsSelected` | `public bool IsSelected` — whether this is the active output route; always `false` where no backend reports it |
| `PlatformAudio.PlayoutPreference` | `public IReadOnlyList<AudioDeviceKind> { get; set; }` — ranked automatic routing policy, default `Bluetooth > WiredHeadset > Speaker > Earpiece`; setter throws `ArgumentNullException`, or `ArgumentException` on `Unknown` / duplicates; a `SetPlayoutDevice` override shadows it until cleared |
| `PlatformAudio.ClearPlayoutDeviceSelection` | `public void ClearPlayoutDeviceSelection()` — clears the sticky `SetPlayoutDevice` override so `PlayoutPreference` applies again; no-op where no override can exist (desktop, iOS, Android < API 31) |
| `PlatformAudio.DevicesChanged` | `public event Action<IReadOnlyList<AudioDevice>, IReadOnlyList<AudioDevice>>` — `(playout, recording)`, raised on the Unity main thread |

#### Changed

No signature changed; these are behavior/contract changes on existing members.

- **Disposed contract**: every public member now throws `ObjectDisposedException` after `Dispose`; `Dispose` is idempotent; subscribing to / unsubscribing from `DevicesChanged` stays safe post-dispose.
- **`GetDevices()`** now goes through the platform routing backend instead of straight to the FFI. iOS returns the audio session's current output route with `Kind`/`IsSelected`; Android 12+ returns the available communication devices. Previously a single placeholder entry per list on all mobile. `SetPlayoutDevice(uint)` / `SetRecordingDevice(uint)` resolve indices against that new list.
- **`SetPlayoutDevice(string)` / `(uint)`** now go through the routing backend as a sticky override of `PlayoutPreference` (the earlier `SelectOutput` draft was folded into this existing member instead of adding a second verb). Desktop: unchanged — FFI device selection, unknown GUID throws `InvalidOperationException` as before. Android 12+: was a no-op, now pins the device as the communication device — dropped when the device disappears, deferred while no `Room` is connected — and an unknown GUID throws `InvalidOperationException`. iOS and Android < API 31: still no effect, now logged as a warning. Both samples dropped their `SetPlayoutDevice(0)`-at-startup call: on Android 12+ it would pin whichever device the OS lists first and shadow the ranking for the whole session.
- **`StartRecording()`** gained an iOS microphone-permission gate, so it now yields across frames on iOS and throws `InvalidOperationException` when the user denies. It also re-asserts the routing policy after capture starts, and switches the iOS session to its recording state.
- **`StopRecording()`** returns the iOS session to its idle, music-friendly state.
- **Redundant `StartRecording()` / `StopRecording()` calls** are left to the platform ADM, which ignores a start while already recording and a stop while idle. The SDK deliberately adds no guard of its own: the request-side flag it keeps to drive the iOS session-state machine can diverge from the ADM's real state (e.g. after the platform stops the capture on an iOS interruption), so gating on it could refuse a needed restart. A public `IsRecording` query was tried in an earlier revision and dropped for the same reason; both samples' `PlatformAudioController` just call through instead of keeping their own `_isRecording` bookkeeping.
- **Constructor**: iOS session is now app-owned / manual mode with lifecycle observers. Construction takes no call audio session on any platform; the session follows `Room` connections (next bullet).
- **Call audio session follows `Room` connections** (new behavior, no new API): `PlatformAudio` holds the platform's call audio session — on iOS WebRTC's VPIO unit, on Android 12+ `MODE_IN_COMMUNICATION` plus the output route pin — while at least one `Room` is connected and releases it when the last one disconnects. Taken and released inside the room's connection-state transition, before `Room.Connected` / `Room.Disconnected` reach app handlers, for a local `Disconnect()` and a server-side disconnect alike; a reconnect in progress still counts as a call. An instance created while a room is already connected takes the session immediately; one created outside a call sits idle (music-friendly iOS state, platform routing on Android). Plumbing is `internal` (`Room.ConnectedRoomCount` / `Room.ConnectedRoomCountChanged`, maintained in `Room.SetConnectionState`). There is deliberately no manual override: a room that carries no audio simply should not have a `PlatformAudio` alive.
- **`Room.Disconnect()` / `Dispose()`** (from #382, merged in): a local disconnect is now reported like a server-side one — `ConnectionStateChanged`, `Disconnected`, `DisconnectedWithReason` with `DisconnectReason.ClientInitiated` — synchronously with the room's handles still live, so one teardown handler covers both; `IsConnected` is true from `Connected` on. Details and tests in #382.
- **`Dispose()`**: refcounted restore of the pre-LiveKit iOS audio session once the last instance goes away.
- Doc correction, not a behavior change: `StartRecording`'s old claim that recording auto-starts on construction was never true.

#### Removed

Nothing relative to `main` / 2.0.0. No public type, member, or overload was removed, and no signature, parameter, or return type changed. (`PlatformAudio.SetSessionAudioEnabled` and `PlatformAudio.IsRecording`, both added in earlier revisions of this PR, were dropped again before merge — the former in favor of the automatic behavior above, the latter because it could misreport the ADM's real state; neither ever shipped.) `AudioDevice` is a struct that grew two fields — source-compatible; only relevant when linking a precompiled assembly, which this source-distributed UPM package does not do.

#### New integration requirement

Android routing needs `<uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />` in the app's `AndroidManifest.xml` (both sample manifests updated).

#### Proposed semver

**Minor — 2.1.0** (from `package.json` 2.0.0). The surface is purely additive, so existing code compiles and runs unchanged. The behavior changes are real but confined to `PlatformAudio` — worth prominent release notes (the `MODIFY_AUDIO_SETTINGS` requirement, the call audio session now following `Room` connections — nothing to gate in app code, and a `PlatformAudio` kept alive next to an audio-less room now takes the call session for it — the iOS `StartRecording` yield, `SetPlayoutDevice` now routing on Android 12+ — drop select-device-0-at-startup calls — and, via #382, `Room.Disconnected` firing for local disconnects, so teardown handlers must tolerate running on hang-up) rather than a major bump.


Runtime/Plugins/iOS/LiveKitAudioSession.mm
```diff
@@ -15,55 +15,590 @@
...
+        if (type.unsignedIntegerValue == AVAudioSessionInterruptionTypeEnded) {
+            LiveKit_ScheduleSessionRecovery();
+        }
+    }];
+    [center addObserverForName:AVAudioSessionRouteChangeNotification
+                        object:nil
+                         queue:[NSOperationQueue mainQueue]
+                    usingBlock:^(NSNotification* note) {
+        if (!s_liveKitConfigured) {
+            return;
```
xianshijing-lk, line 354: should we remove the observer in this case so the rest of the code will be run after s_liveKitConfigured is true ? 
or they don't matter any more ?

davidliu:
Probably something to note for android as a later TODO, since you're working with communication mode:

Android MODE_IN_COMMUNICATION drops if no audio playback or recording is done for 6 seconds. The most common repro case would be if the client joins an empty room with no mic turned on, the communication will drop unexpectedly.
https://issuetracker.google.com/issues/209493718

The workaround Android uses is to just play zeroed audio during these silent moments.
https://github.com/livekit/client-sdk-android/pull/363

Samples~/Agents/Assets/Runtime/Agent/PlatformAudioController.cs
```diff
@@ -1,35 +1,88 @@
...
+    // Whether a remember pass has ever swept the scene; gates the adopt-loops fallback
+    // in RestartAudibleSources to the very first switch.
+    bool _sceneSwept;
 
+    public bool IsInitialized => _platformAudio != null;
     public bool IsPublished { get; private set; }
 
+    public PlatformAudioController(string trackName, AudioProcessingOptions audioOptions)
+    {
+        _trackName = trackName;
```
xianshijing-lk, line 50: why do we need to take a custom trackName from the app ?

Samples~/Agents/Assets/Runtime/Agent/LiveKitAgentSession.cs
```diff
@@ -81,7 +85,7 @@ IEnumerator Connect()
         // Create the WebRTC ADM before connecting. The SDK only wires automatic speaker
         // playout for remote tracks to a PlatformAudio that already exists at connect time;
         // initializing it after Connect leaves remote (agent) audio silent.
-        _audio = new PlatformAudioController();
+        _audio = new PlatformAudioController(MicTrackName, AudioProcessingOptions.Default);
```
xianshijing-lk, line 88: Do you think we will allow changing the audio processing on the fly ? 

And I wonder if this AudioProcessingOptions should go to startCapture() instead for the long term

Samples~/Meet/Assets/Runtime/PlatformAudioController.cs
```diff
@@ -0,0 +1,398 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using LiveKit;
+using LiveKit.Proto;
+using UnityEngine;
+
+// Drives the duplex platform audio (WebRTC ADM): captures the default microphone with
```
xianshijing-lk, line 9: is this PlatformAudioController.cs the same as the one at Agents/Runtime/Agent ? 

If yes, any chance that we can share the code ?

Samples~/Agents/Assets/Plugins/Android/AndroidManifest.xml
```diff
@@ -7,6 +7,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.CAMERA" />
+    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
```
davidliu, line 10: Is this something we can bake into the library rather than require consumers to add themselves? Not sure on how Unity packages their libraries.

Alternatively, adding a permission check to AndroidRouteController for MODIFY_AUDIO_SETTINGS might be a good idea to have regardless:
 `Android.Permission.HasUserAuthorizedPermission("android.permission.MODIFY_AUDIO_SETTINGS")`

Runtime/Scripts/Audio/PlatformAudio.cs
```diff
@@ -241,6 +591,7 @@ public void SetRecordingDevice(string deviceId)
         /// </exception>
         public void SetPlayoutDevice(uint index)
         {
+            ThrowIfDisposed();
             var (_, playout) = GetDevices();
```
davidliu, line 566: I think this should be GetDevicesViaFFI? GetDevices used to be grabbing from FFI previously, but now it's grabbing the platform devices, which mean if the GUIDs aren't exactly the same, this'll be unsafe.

Alternatively, keep GetDevices under the old retrieve from FFI behavior to ensure that consumers don't break, and introduce the new behavior under a new method.

Samples~/Meet/Assets/Scenes/MeetApp.unity
```diff
@@ -853,6 +853,134 @@ Canvas:
   m_SortingLayerID: 0
   m_SortingOrder: 0
   m_TargetDisplay: 0
+--- !u!1 &1060469597
```
MaxHeimbrock, line 856: I should probably check what changed in the scene. I don't think this should be commited.

PR #381, opened 2026-09-04
AEC in Unity Audio

MaxHeimbrock:
### Background

Unity Audio had no echo cancellation, only `PlatformAudio` had AEC.

There is a new `PlayoutReference` which captures the final game audio from the `AudioListener` in the scene. That is the far-end input for the AEC

With audio processing enabled, the `RtcAudioSource` (which represents our microphone) creates an `AudioProcessor`, which itself holds the `AudioProcessingModule` (APM) wrapping the native APM from WebRTC through FFI. We already had all the APM stuff exposed through FFI anyways, so this is not in this PR.

The `AudioProcessor` receives the mic input from `RtcAudioSource` and the scene audio from the `PlayoutReference plus an occasional new delay hint  (every 2s) and does its processing using its `AudioProcessingModule`, before sending the audio input to the native side. 

With audio processing disabled, the input audio still goes straight to the native side.

image

### API surface

| Type | Member | Change |
|---|---|---|
| `AudioProcessingOptions` | `bool PreferHardware` | Unchanged. Documented as `PlatformAudioSource` only. |
| `MicrophoneSource` | `MicrophoneSource(string, GameObject, AudioProcessingOptions)` | Added constructor. |
| `RtcAudioSource` | `protected RtcAudioSource(RtcAudioSourceType, AudioProcessingOptions)` | Added constructor. |
| `RtcAudioSource` | `protected RtcAudioSource(RtcAudioSourceType, uint, uint, AudioProcessingOptions?)` | Added constructor. |
| `RtcAudioSource` | `bool AudioProcessingEnabled { get; }` | Added. |
| `RtcAudioSource` | `protected void ResetAudioProcessing()` | Added. |
| `PlayoutReference` | `MonoBehaviour` | New component. No public members; the SDK adds it to the active `AudioListener` when echo cancellation is on.|

Nothing removed or changed in signature. `AudioProcessingOptions` keeps its three stage flags; The FFI wrapper `AudioProcessingModule` is `internal`; the public entry point is `MicrophoneSource` / `RtcAudioSource` with `AudioProcessingOptions`.

## Verification

Tested manually on MacOS, iOS and Android


Samples~/Meet/Assets/Scenes/MeetApp.unity
```diff
@@ -902,7 +902,7 @@ MonoBehaviour:
   videoTrackParent: {fileID: 2128321498}
   participantTilePrefab: {fileID: 4315784896331113596, guid: bec493bbc3d574c07b5bbf8dd2be26b3, type: 3}
   frameRate: 30
-  usePlatformAudio: 1
+  usePlatformAudio: 0
```
MaxHeimbrock, line 905: Sets Unity audio as the default

Runtime/Scripts/Audio/Processing/AudioProcessor.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// </summary>
+        public void RequestReset()
+        {
+            Interlocked.Exchange(ref _captureResetRequested, 1);
+            Interlocked.Exchange(ref _referenceResetRequested, 1);
+        }
+
+        // Periodic main-thread upkeep: re-attach the reference after scene or device changes and
+        // refresh the delay hint (iOS reports zero session latency until the session is active).
+        private IEnumerator Maintenance(int generation)
```
MaxHeimbrock, line 142: This is a choice. To me it does not sound harmful to ensure the correct setup.

Runtime/Scripts/Audio/RtcAudioSource.cs
```diff
@@ -174,45 +210,59 @@ private void OnAudioRead(float[] data, int channels, int sampleRate)
...
+            // Optional processing stage: the block is re-chunked into 10 ms frames, run through the
+            // module and delivered to SendFrame one chunk at a time via SendProcessedFrame.
+            if (_processor != null && _processor.TryProcessCapture(data, channels, sampleRate))
+                return;
 
             // Each captured frame gets its own backing buffer so the native encoder can safely
             // consume it asynchronously after request.Send() returns.
             var frameData = new NativeArray<short>(data.Length, Allocator.Persistent);
+            for (int i = 0; i < data.Length; i++)
+                frameData[i] = PcmConvert.FloatToS16(data[i]);
```
MaxHeimbrock, line 245: We are reusing the newly introduced `PcmConvert.FloatToS16` here instead of the old local `FloatToS16`.

Runtime/Scripts/Audio/Processing/AudioProcessingDelayHint.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public static int EstimateMs()
+        {
+            var config = AudioSettings.GetConfiguration();
+            return EstimateMs(config.dspBufferSize, config.sampleRate, PlatformLatencyMs());
+        }
+
+        internal static int EstimateMs(int dspBufferSize, int sampleRate, double deviceLatencyMs)
+        {
+            var blockMs = sampleRate > 0 ? dspBufferSize * 1000.0 / sampleRate : 0.0;
```
xianshijing-lk, line 53: [question](https://pullpo.io/cc?l=question) 
any chance the sampleRate can be 0 in some corner cases ?

Runtime/Scripts/Audio/Processing/AudioProcessingDelayHint.cs
```diff
@@ -0,0 +1,78 @@
...
+    internal static class AudioProcessingDelayHint
+    {
+#if UNITY_IOS && !UNITY_EDITOR
+        [DllImport("__Internal")] private static extern double LiveKit_AudioSessionOutputLatency();
+        [DllImport("__Internal")] private static extern double LiveKit_AudioSessionInputLatency();
+        [DllImport("__Internal")] private static extern double LiveKit_AudioSessionIOBufferDuration();
+#endif
+
+        internal const int MinDelayMs = 0;
+        internal const int MaxDelayMs = 500;
```
xianshijing-lk, line 30: can you explain where this |MaxDelayMs| comes from ? why we need to cap it with 500ms ?

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        /// <summary>
+        /// Processes the near-end (capture) stream in place. <paramref name="byteCount"/> is bytes,
+        /// not samples — the buffer is interleaved int16. Returns the FFI error, or null on success.
+        /// </summary>
+        public string ProcessStream(IntPtr dataPtr, int byteCount, int sampleRate, int channels)
+        {
+            if (_disposed) throw new ObjectDisposedException(nameof(AudioProcessingModule));
+
+            using var request = FFIBridge.Instance.NewRequest<ApmProcessStreamRequest>();
```
xianshijing-lk, line 98: can you check the rust side how this ApmProcessStreamRequest is handled ?

Rust is using a Tokio runtime to handle all the requests from client, and use a thread pool under the hood to handle them. That brings big challenges with handling audio callbacks every 10ms, and sometimes it could take >20ms to handle audio callbacks.
That is why I introduced a dedicated tokio runtime for the audio callback, and tune the thread priority to high there.

maybe you should make sure this NewApmRequest is handled by that audio runtime ?

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// </summary>
+        private static readonly int[] NativeSampleRates = { 8000, 16000, 32000, 48000 };
+
+        private readonly FfiHandle _handle;
+        private bool _disposed;
+
+        /// <summary>The native handle id, for diagnostics.</summary>
+        public ulong Handle => (ulong)_handle.DangerousGetHandle();
+
+        public AudioProcessingModule(
```
xianshijing-lk, line 42: what will happen if device is unplugged during streaming, and the new device comes with a different sample rate ? do we shutdown the old stream and create a new one + new APM ?

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
xianshijing-lk: btw, I wonder if it is actually easier to use uniffi to pass the APM rather than protobuf base ffi. 

and the uniffi can be sync ?

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
```diff
@@ -0,0 +1,167 @@
...
+            using var response = request.Send();
+            FfiResponse res = response;
+            return ErrorOrNull(res.ApmProcessStream?.Error);
+        }
+
+        /// <summary>
+        /// Processes the far-end (render) reference stream in place. Same buffer contract as
+        /// <see cref="ProcessStream"/>.
+        /// </summary>
+        public string ProcessReverseStream(IntPtr dataPtr, int byteCount, int sampleRate, int channels)
```
xianshijing-lk, line 115: same question here, wonder if it is easier to do it with uniffi.

OK to defer to a follow up if that is too complicated

xianshijing-lk (commented):
The review is still ongoing, I am posting what I have now.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
```diff
@@ -0,0 +1,167 @@
...
+            delay.DelayMs = delayMs;
+
+            using var response = request.Send();
+            FfiResponse res = response;
+            return ErrorOrNull(res.ApmSetStreamDelay?.Error);
+        }
+
+        private static string ErrorOrNull(string error) => string.IsNullOrEmpty(error) ? null : error;
+
+        public void Dispose()
```
xianshijing-lk, line 152: Is this must call ? can you add document if it is. 

Otherwise, I think we'd better handle the cases that our developers could forget calling Dispose(), and we handle it more gratefully.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
```diff
@@ -0,0 +1,167 @@
...
+            using var response = request.Send();
+            FfiResponse res = response;
+            return ErrorOrNull(res.ApmProcessReverseStream?.Error);
+        }
+
+        /// <summary>
+        /// Seeds the render/capture delay. AEC3 runs its own correlation estimator, so this is a
+        /// convergence hint rather than a hard alignment. Returns the FFI error, or null on success.
+        /// </summary>
+        public string SetStreamDelayMs(int delayMs)
```
xianshijing-lk, line 136: what thread will this SetStreamDelayMs() be called ? if this can be called rather than the main thread, is it thread safe ?

Runtime/Scripts/Audio/RtcAudioSource.cs
```diff
@@ -328,6 +391,7 @@ protected virtual void Dispose(bool disposing)
                 }
                 _pendingFrameData.Clear();
             }
+            _processor?.Dispose(disposing);
             Handle?.Dispose();
             _disposed = true;
```
xianshijing-lk, line 396: quick question, should _disposed = true; be moved up right after if (_disposed) return; ? 
like 
if (_disposed) return;
_disposed = true;

// Then it will dispose all the components. 

In case the second Dispose() is called again, the disposing can happen only once. 

Ignore it if Dispose() can be called by only one thread.

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
```diff
@@ -0,0 +1,155 @@
...
+            foreach (var listener in listeners)
+            {
+                if (listener.isActiveAndEnabled) return listener;
+            }
+            return null;
+        }
+
+        private void OnEnable()
+        {
+            _listener = GetComponent<AudioListener>();
```
xianshijing-lk, line 97: curiously, so _listener is created only once ? 

I wonder if we should do it in RefreshDeliveryState() as well ?

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
```diff
@@ -0,0 +1,155 @@
...
+        internal static void Acquire()
+        {
+            _consumers++;
+            EnsureAttached();
+        }
+
+        /// <summary>Main thread. The component stays on the listener; it is inert without consumers.</summary>
+        internal static void Release()
+        {
+            if (_consumers > 0) _consumers--;
```
xianshijing-lk, line 63: what will happen if _consumers == 0 ?

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
```diff
@@ -0,0 +1,155 @@
...
+        private void Update()
+        {
+            RefreshDeliveryState();
+        }
+
+        // Listener state and the output rate are Unity APIs; sample them here for the audio thread.
+        private void RefreshDeliveryState()
+        {
+            _sampleRate = AudioSettings.outputSampleRate;
+            _deliver = _listener != null && _listener.isActiveAndEnabled;
```
xianshijing-lk, line 122: that might be related to the consumer_ question should _deliver be true if _consumer is 0 ?

Runtime/Scripts/Audio/RtcAudioSource.cs
```diff
@@ -171,48 +207,68 @@ public virtual void Stop()
...
+            // Optional processing stage: the block is re-chunked into 10 ms frames, run through the
+            // module and delivered to SendFrame one chunk at a time via SendProcessedFrame.
+            if (_processor != null && _processor.TryProcessCapture(data, channels, sampleRate))
+                return;
+
+            if (_muted) return;
 
             // Each captured frame gets its own backing buffer so the native encoder can safely
             // consume it asynchronously after request.Send() returns.
             var frameData = new NativeArray<short>(data.Length, Allocator.Persistent);
```
xianshijing-lk, line 243: Curiously, do you know why we should allocate a new blocking buffer for each frame ? 

Is OnAudioRead thread safe ? and I wonder if it is viable to re-use the buffer rather than allocate a new one for each frame

Runtime/Scripts/Audio/Processing/AudioProcessingDelayHint.cs
MaxHeimbrock: That mirrors the max delay in libwebrtc: https://github.com/webrtc-sdk/webrtc/blob/m150_release/modules/audio_processing/audio_processing_impl.cc#L1645C15-L1645C18

So any value outside of 0 - 500ms is rejected by libwebrtc and throws a warning.

The comment there also says it is a bit arbitrary, but technically the AEC should only take the hint and then converge to the real delay. 

Talking to AI a bit about how the libwebrtc actually converges, that also only takes 0.5s of history into account to match the audio output signal in the mic input again, testing five overlapping windows:

```
  ┌────────┬───────────────────────┐
  │ Filter │ Covers echo delays of │
  ├────────┼───────────────────────┤
  │ 0      │ 0 to 128 ms           │
  ├────────┼───────────────────────┤
  │ 1      │ 96 to 224 ms          │
  ├────────┼───────────────────────┤
  │ 2      │ 192 to 320 ms         │
  ├────────┼───────────────────────┤
  │ 3      │ 288 to 416 ms         │
  ├────────┼───────────────────────┤
  │ 4      │ 384 to 512 ms         │
  └────────┴───────────────────────┘
```

Runtime/Scripts/Audio/Processing/AudioProcessingDelayHint.cs
MaxHeimbrock: Yes, I checked again and if you disable the Unity audio system, the sample rate is reported as 0. In this case we have the guard, but there was another case where this could have caused issues. I pushed a new commit to guard against this as well: https://github.com/livekit/client-sdk-unity/pull/381/changes/0a23b8e41c3eb0d887efcedaeb72266b75fae4aa

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
MaxHeimbrock: The runtime in this case is the Unity audio thread, as this is the calling thread and it is running in sync, so the handling should run on the same thread. The Unity audio thread usually maps to an OS wide audio thread with high priority. 

I would like to wait with a uniffi implementation to not block shipping this feature.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
MaxHeimbrock: I don't want to block this, but in the future this will be planned of course.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
MaxHeimbrock: Sadly the Unity Audio path is not robust against device changes at runtime. We still need to implement a robust system around it, not only for AEC but to move the mic around, playout mostly works. 

I would postpone this, I have tickets for that and when this AEC is proven to work also for a range of customers I see the Unity audio path actually as the better one over Platform Audio, just we have to put more work into it fixing the mic moving devices.

MaxHeimbrock: And the exact answer is we keep the old input stream and APM but the stream is silent now. We should recreate at least the stream and I assume also the APM when we add robust device switching.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
MaxHeimbrock: The class is completely internal and the developers don't have to deal with it. The Dispose() is handled both in the finalizer `~AudioProcessingModule` below this dispose and the `RtcSource` disposes it on its own disposition or finalization. So even if the developers forget to dispose the `RtcSource` it will eventually be cleaned up C# style.

Runtime/Scripts/Audio/Processing/AudioProcessingModule.cs
MaxHeimbrock: It is called in a coroutine which runs on the Unity main thread and only there (the maintenance coroutine). The APM is internal, so developers can not call it, even though the method signature is public. I assume it counts therefore as thread safe.

Runtime/Scripts/Audio/RtcAudioSource.cs
MaxHeimbrock: It is still not guaranteed thread safe if we don't use `Interlocked.Exchange` to set the bool. In practice, the disposing will not be called from anything else than the main thread, because it also stops the Unity microphone in `Stop()`, which is a Unity API only allowed to be called from the main thread.

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
MaxHeimbrock: It is not created once. The AudioListener should already be a component in the scene. In Unity, you basically have one main virtual camera in every scene, and that usually comes with one virtual microphone attached. The virtual microphone is registering all the audio sources in the scene and is called an `AudioListener`. The Unity audio system kind of enforces you to have only one, throwing a warning if you have more (while you can add more cameras without `AudioListener` to the scene). This virtual microphone sends the sounds recorded in the scene straight to the selected physical playout device.

So using that singleton pattern, in this class we just find it with `FindActiveListener` and attach our script to it to listen in to all the sounds going to the speaker, sending that as input to the echo cancellation.
 
Screenshot 2026-09-21 at 15 23 28

MaxHeimbrock: In the maintenance loop we `EnsureAttached()` every 2s, if for example the camera got recreated through an additive scene change or something else.

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
MaxHeimbrock: The `Update()` triggers a check if we have consumers, in each frame. 

If we don't have any consumers, the audio taken from the `AudioListener` in `OnAudioFilterRead` is not delivered through the `AudioRead` invocation, because `_deliver` is false.

Runtime/Scripts/Audio/Processing/PlayoutReference.cs
MaxHeimbrock: See https://github.com/livekit/client-sdk-unity/pull/381#discussion_r4062592328 above.

Runtime/Scripts/Audio/RtcAudioSource.cs
MaxHeimbrock: Very good point. I checked again and was able to drop the new NativeArray in exchange for a reusable short array.

I tested it on device and it still works. Also I profiled the Android audio thread CPU usage and it is still stable, a bit higher (20% - 30% CPU utilization) than without echo cancellation (10% - 20% CPU utilization).

xianshijing-lk (approved):
lgtm

PR #382, opened 2026-09-08
Report client-initiated disconnects through Room.Disconnected

MaxHeimbrock:
### Background

`Room.Disconnected` / `DisconnectedWithReason` were only raised for server-side disconnects. A local `Disconnect()` (and `Dispose()`) sent the FFI request and ran the cleanup, which unsubscribes the room from FFI events synchronously, so the `ClientInitiated` Disconnected event the Rust core emits during `close()` never reached the app. 

Related gap: `ConnectionState` never left `ConnDisconnected` after a connect, because the Rust core records the Connected transition during `connect`, before this wrapper subscribes to room events. `IsConnected` was therefore false on a connected room. The two `[Ignore("Known issue")]` tests in `RoomTests` covered exactly these two gaps, which are now enabled again: [link](https://github.com/livekit/client-sdk-unity/pull/382/changes#diff-3088cbdb2fb9ba86ef330dfe6be9e125c1bb2a901e5594c4772fc35a64a150cbL235)

### Changes

- `Room.Disconnect()` and `Dispose()` now raise `ConnectionStateChanged(ConnDisconnected)`, `Disconnected` and `DisconnectedWithReason` with `DisconnectReason.ClientInitiated`. 
- All disconnect paths (local, core `Disconnected` event, panic) share one `Teardown(reason, closeFfiRoom)`: mark disposed, unsubscribe from FFI events, raise events, release handles in a `finally`. Re-entrant or repeated `Disconnect()` is a no-op.
- `OnConnect` records `ConnConnected` and raises `ConnectionStateChanged`, so `IsConnected` is true once `Connected` is raised. The core's own copy is deduplicated. `OnConnect` also resets the disposed flag, so a reconnected Room gets a full lifecycle.
- `ConnectInstruction` completes in a `finally`, so a throwing `OnConnect` handler no longer hangs the awaiting code.

- Meet sample: `OnEndCall` only calls `Disconnect()`; `OnDisconnected` owns the teardown for hang-up, server-side disconnect and `OnDestroy`. Server-side disconnects previously left tracks and the "connected" UI behind.
- Tests: `ConnectionState_IsConnected` and `Disconnect_TriggersEvent` un-ignored. New PlayMode and EditMode tests cover the single `Connected` report, the client-initiated disconnect (live handles, re-entrant `Disconnect()`, `Dispose()`, throwing handler) and re-entrant disconnects keeping the server reason.

#### Behavior change for release notes

Every existing `Disconnected` handler now also runs on the app's own `Disconnect()` / `Dispose()`. A handler that is not idempotent with the code around the hang-up call will tear down twice. The Meet sample is updated to the single-handler shape in this PR; the Agents sample does not subscribe to `Disconnected` and is unaffected.

### DIFF
--- a/Runtime/Scripts/Core/Room.cs
+++ b/Runtime/Scripts/Core/Room.cs
@@ -113,6 +113,11 @@ namespace LiveKit
     {
         internal FfiHandle RoomHandle = null;
         private bool _disposed = false;
+        // Set once the disconnect has been reported to the app (Disconnected /
+        // DisconnectedWithReason raised) by whichever path got there first: the remote
+        // Disconnected event, a panic, or a local Disconnect(). Keeps the report to
+        // exactly one per room, also when a handler calls Disconnect() re-entrantly.
+        private bool _disconnectReported;
         private readonly Dictionary<string, RemoteParticipant> _participants = new();
         private StreamHandlerRegistry _streamHandlers = new();
 
@@ -190,9 +195,19 @@ namespace LiveKit
             return instruction;
         }
 
+        /// <summary>
+        /// Disconnects from the room. Like the other LiveKit SDKs, a local disconnect is
+        /// reported through <see cref="ConnectionStateChanged"/>, <see cref="Disconnected"/>
+        /// and <see cref="DisconnectedWithReason"/> with
+        /// <see cref="DisconnectReason.ClientInitiated"/> — synchronously, before the
+        /// room's resources are released — so one teardown path covers hang-ups and
+        /// server-side disconnects alike. <see cref="Dispose"/> goes through here too.
+        /// Calling this from a Disconnected handler, or on a room that already reported
+        /// its disconnect, is a no-op.
+        /// </summary>
         public void Disconnect()
         {
-            if (_disposed || RoomHandle == null)
+            if (_disposed || _disconnectReported || RoomHandle == null)
                 return;
             var (response, _) = FFIBridge.Instance.SendDisconnectRequest(this);
             using (response)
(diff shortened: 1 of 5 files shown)


Runtime/Scripts/Core/Room.cs
```diff
@@ -613,6 +684,15 @@ internal void OnConnect(ConnectCallback info)
...
             }
 
+            // The core recorded this transition during connect, before this room was
+            // subscribed to its events (its own copy drains later and is dropped as a
+            // repeat, see SetConnectionState); recording it here makes IsConnected true
+            // from the moment Connected is raised.
+            SetConnectionState(ConnectionState.ConnConnected);
+            // A ConnectionStateChanged handler may already have disconnected the room;
+            // Connected is not raised on a torn-down room.
+            if (_disposed)
```
xianshijing-lk, line 694: if _disposed is true, while the connection state just gets set to ConnConnected, do they get out of sync then ? 

Should we do some cleanup here ?

Samples~/Meet/Assets/Runtime/MeetManager.cs
```diff
@@ -142,17 +142,14 @@ private void Update()
...
-        }
+        // streams, and their backing GPU/native resources allocated. Disconnect reports
+        // back through OnDisconnected, which tears the call down; CleanUpAllTracks after
+        // it covers anything created without a room.
+        _room?.Disconnect();
         CleanUpAllTracks();
         _webCamTexture?.Stop();
         _platformAudioSource?.Dispose();
         _platformAudio?.Dispose();
-        _room?.Disconnect();
```
xianshijing-lk, line 155: any reason to call _room?.Disconnect(); before shutting down the sources ? 

and if _room?.Disconnect() is async, I wonder if we should wait until everything gets shutdown to clean up everything

Runtime/Scripts/Core/Room.cs
MaxHeimbrock: The only way this case `if (_disposed)` is hit is when the handler registered to OnConnectionStateChanged is disconnecting the room, because all other calls between setting `_disposed = false` in line 663 and the `SetConnectionState` in line 691 are sync calls without any handlers.

If the handler on OnConnectionStateChanged would disconnect the room, then the `ConnectionState` will already be set to disconnected once line 694 is reached and the value is in sync with `_disposed` again. In that case the `Connected` event should be skipped, so the gate is correct here.

Samples~/Meet/Assets/Runtime/MeetManager.cs
MaxHeimbrock: The removed `_room?.Disconnect();` at the end was just a duplication. Usually the removed line 148 was calling `_room.Disconnect()` already at the top and it stays this way in the PR. I don't see an issue doing these things in parallel. 

The room disconnect message should be independent from the other resources being torn down, except for PlatformAudio, which is safer to not dispose while the room is still active in my opinion.

xianshijing-lk (approved):

